PR #42, opened 2021-11-16
Region Domain: MRU Cache

LinerSu:
This PR includes the following updates:
(1) Add a new variable domain that is used to represent the base address of an allocated object symbolically. The Hasse diagram is:
```
                   Top
        /     /     |     \   
...  var1   var2   ...  varn  ...
        \      \    |     /     
                   Bot
```
(2) Enable MRU cache in region domain. The basic process is during `load_ref / store_ref`:
If load/store a data for a object by reference within a region:
- Check whether the object is an MRU object
    - If not, cache missed; commit cache to base domain; populate new MRU object into cache domain
- Perform strong read / update

### DIFF
--- a/include/crab/domains/term/term_operators.hpp
+++ b/include/crab/domains/term/term_operators.hpp
@@ -13,16 +13,19 @@ namespace term {
 
 class term_operator_t {
   uint32_t m_value;
-  boost::string_ref m_name;
+  bool m_reserved;
 public:
-  constexpr term_operator_t(uint32_t value)
-    : m_value(value), m_name(boost::string_ref()) {}
-  term_operator_t(uint32_t value, boost::string_ref name);
+  // used for defining constexpr crab operators
+  constexpr term_operator_t(uint32_t value, bool is_reserved = true)
+    : m_value(value), m_reserved(is_reserved) {}
+  // used for defining new user operators
+  static term_operator_t make_operator(uint32_t value);  
   ~term_operator_t() = default;
   constexpr operator uint32_t() const { return m_value; }
   constexpr uint32_t value() const { return m_value; }
-  boost::string_ref name() const { return m_name; }
-  constexpr static uint32_t first_nonreserved_value() { return 100;}
+  constexpr bool is_reserved() const { return m_reserved;}
+  constexpr static uint32_t first_nonreserved_value() { return 100;}  
+  friend crab::crab_os &operator<<(crab::crab_os &o, term_operator_t op);
 };
 
 constexpr term_operator_t TERM_OP_ADD(0);
@@ -44,76 +47,6 @@ constexpr term_operator_t TERM_OP_LSHR(15);
 constexpr term_operator_t TERM_OP_ASHR(16);
 constexpr term_operator_t TERM_OP_FUNCTION(17);  
   
-inline crab::crab_os &operator<<(crab::crab_os &o, term_operator_t op) {
-
-  // User-defined operator
-  if (const char* name = op.name().data()) {
-    o << name;
-    return o;
-  }
-
-  // Crab predefined operators
-  switch (op) {
-  case TERM_OP_ADD:
-    o << "+";
-    break;
-  case TERM_OP_SUB:
-    o << "-";
-    break;
-  case TERM_OP_MUL:
-    o << "*";
-    break;
-  case TERM_OP_SDIV:
-    o << "/";
-    break;
-  case TERM_OP_UDIV:
-    o << "/_u";
-    break;
-  case TERM_OP_SREM:
-    o << "%";
-    break;
-  case TERM_OP_UREM:
-    o << "%_u";
-    break;
-  case TERM_OP_NOT:
-    o << "not";
-    break;
-  case TERM_OP_BAND:
-    o << "band";
-    break;
-  case TERM_OP_BOR:
-    o << "bor";
-    break;
-  case TERM_OP_BXOR:
-    o << "bxor";
-    break;
-  case TERM_OP_AND:
-    o << "&";
-    break;
-  case TERM_OP_OR:
-    o << "|";
-    break;
-  case TERM_OP_XOR:
-    o << "^";
-    break;
-  case TERM_OP_SHL:
-    o << "<<";
-    break;
-  case TERM_OP_LSHR:
-    o << ">>_l";
-    break;
-  case TERM_OP_ASHR:
-    o << ">>_r";
-    break;
-  case TERM_OP_FUNCTION:
-    o << "uf";
-    break;
-  default:
-    CRAB_ERROR("unexpected binary operation ", op);
-  }
-  return o;
-}
-
 /* Convert between Crab operators and term domain uninterpreted functors */
 term_operator_t conv2termop(arith_operation_t op);
 term_operator_t conv2termop(bitwise_operation_t op);
--- a/include/crab/domains/uf_domain.hpp
+++ b/include/crab/domains/uf_domain.hpp
@@ -638,9 +638,8 @@ public:
 
   // Precondition:
   // - value must be > term_operator_t::first_nonreserved_value()
-  // Note that boost::string_ref is a non-owning reference to a string.
-  static term::term_operator_t make_uf(uint32_t value, boost::string_ref name) {
-    term::term_operator_t op(value, name);
+  static term::term_operator_t make_uf(uint32_t value) {
+    auto op = term::term_operator_t::make_operator(value);
     return op;
   }
 
--- a/lib/term_operators.cpp
+++ b/lib/term_operators.cpp
@@ -4,14 +4,84 @@ namespace crab {
 namespace domains {
 namespace term {
 
-term_operator_t::term_operator_t(uint32_t value, boost::string_ref name)
-  : m_value(value), m_name(name) {
-  if (m_value < first_nonreserved_value()) {
-    CRAB_ERROR("First " + std::to_string(first_nonreserved_value()) +
+term_operator_t term_operator_t::make_operator(uint32_t value) {
+  term_operator_t op(value, false /*user-defined*/);
+  if (op.value() < op.first_nonreserved_value()) {
+    CRAB_ERROR("First " + std::to_string(op.first_nonreserved_value()) +
 	       "numbers are reserved in term_operator_t");
-  }  
+  }
+  return op;
 }
 
+crab::crab_os &operator<<(crab::crab_os &o, term_operator_t op) {
+ 
+  if (!op.is_reserved()) {
+    // User-defined operator      
+    o << "UF_SYM_" << std::to_string(op.value());
+  } else {
+    // Crab predefined operators
+    switch (op) {
+    case TERM_OP_ADD:
+      o << "+";
+      break;
+    case TERM_OP_SUB:
+      o << "-";
+      break;
+    case TERM_OP_MUL:
+      o << "*";
+      break;
+    case TERM_OP_SDIV:
+      o << "/";
+      break;
+    case TERM_OP_UDIV:
+      o << "/_u";
+      break;
+    case TERM_OP_SREM:
+      o << "%";
+      break;
+    case TERM_OP_UREM:
+      o << "%_u";
+      break;
+    case TERM_OP_NOT:
+      o << "not";
+      break;
+    case TERM_OP_BAND:
+      o << "band";
+      break;
+    case TERM_OP_BOR:
+      o << "bor";
+      break;
+    case TERM_OP_BXOR:
+      o << "bxor";
+      break;
+    case TERM_OP_AND:
+      o << "&";
+      break;
+    case TERM_OP_OR:
+      o << "|";
+      break;
+    case TERM_OP_XOR:
+      o << "^";
+      break;
+    case TERM_OP_SHL:
+      o << "<<";
+      break;
+    case TERM_OP_LSHR:
+      o << ">>_l";
+      break;
+    case TERM_OP_ASHR:
+      o << ">>_r";
+      break;
+    case TERM_OP_FUNCTION:
+      o << "uf";
+      break;
+    default:
+      CRAB_ERROR("unexpected term_operator_t op");
+    }
+  }
+  return o;
+}
+  
 term::term_operator_t conv2termop(arith_operation_t op) {
   switch (op) {
   case OP_ADDITION:


include/crab/domains/region_domain.hpp
```diff
@@ -406,6 +408,10 @@ class region_domain final
       }
     }
 
+    // TODO: generalize expand function
+    void expand(base_abstract_domain_t &dom, base_abstract_domain_t &cache,
+                const std::vector<ghost_variables_t> &rgn_list) const {}
```
LinerSu: TODO1: Need to have a new expand function to populate cache.

include/crab/domains/region_domain.hpp
```diff
@@ -561,6 +567,30 @@ class region_domain final
...
+  // left: the base address for cache
+  // right: a region variable to get equivalence class
+  using cache_variable_opt_t =
+      boost::optional<std::pair<base_variable_domain_t, variable_t>>;
+  cache_variable_opt_t m_cache_vars;
+  // key: a reference variable
+  // value: a base variable in the base domain that represents the base address
+  // of the MRU
+  using base_var_map_t =
+      ikos::separate_domain<variable_t, base_variable_domain_t>;
```
LinerSu: The current environment cannot resolve the following code:
```c
p := make_ref(…)
q := make_ref(….)
if (…) {
r := gep_ref(p, …);
} else {
r := gep_ref(q, …)
}
store_ref(r,...)
```
The reason is the domain only maps from one reference variable into a single base address.

include/crab/domains/region_domain.hpp
```diff
@@ -561,6 +567,30 @@ class region_domain final
...
 
+  /** Begin MRU domain **/
+  // The cache abstract domain that capture invariants for MRU object
+  base_abstract_domain_t m_cache_dom;
+  using base_variable_domain_t = variable_domain<base_variable_t>;
+  // left: the base address for cache
+  // right: a region variable to get equivalence class
+  using cache_variable_opt_t =
+      boost::optional<std::pair<base_variable_domain_t, variable_t>>;
+  cache_variable_opt_t m_cache_vars;
```
LinerSu: I use this pair because we need to keep track of information for the cache domain. That is a reference for the MRU object and a region that MRU used.

LinerSu: The solution is not good enough. Hope you give me some suggestions.

include/crab/domains/region_domain.hpp
```diff
@@ -1325,6 +1428,48 @@ class region_domain final
     }
   }
 
+  /* MRU */
+  void commit_cache() {
+    if (!m_cache_dom.is_bottom()) {
+        // 1. join cache and base domain
+        m_base_dom |= m_cache_dom;
+        // 1.a update init
```
LinerSu: I leave this part because I'm not sure whether we need to distinguish variable expanded from the base domain into cache domain.
For instance,
``` c
region_init(@V_1:region(int));
region_init(@V_2:region(int));

/* Cache: _|_, Sum: T */
s1 := make_ref(@V_1:region(int),8,as_0);
(@V_1:region(int),_0:ref) := gep_ref(@V_1:region(int),s1:ref);
/* Cache: _|_, Sum: T */
store_to_ref(@V_1:region(int),_0:ref,3:int32);
// cache missed. Because cache is empty, require populate cache (expand relationships).
// perform strong update
/* Cache: @V_1_copy = 3, Sum: T */
(@V_2:region(int),_1:ref) := gep_ref(@V_1:region(int),s1:ref + 4);
// Cache hit, perform strong update
store_to_ref(@V_2:region(int),_1:ref,5:int32);
/* Cache: {@V_1_copy = 3, @V_2_copy = 5}, Sum: T */
```
The question is whether we need a map to keep track of correspondence: `V_1_copy <-> V_1; V_2_copy <-> V_2` or we just have the same region variables both in cache domain and base domain?

If the answer is yes, we definitely need to update init and count address for copied region variables.

include/crab/domains/union_find_domain.hpp
```diff
@@ -514,6 +514,21 @@ template <class Variable, class Domain> class union_find_domain {
     m_parents.insert({y, rep_x});
   }
 
+  std::vector<variable_t> get_equiv_vars_by_var(const variable_t &x) {
```
LinerSu: I add this method because I need to get all region variables used by a single object.

include/crab/domains/region_domain.hpp
caballa: yes, we should have directly a set of variables (`discrete_domains.hpp`) to keep track of multiple possible base addresses. But can we do something useful when the set is not a singleton?

include/crab/domains/region_domain.hpp
caballa: Just some context that will help @agurfinkel  to follow this but I'm not answering directly the question yet.

The base domain is the underlying numerical domain. So top-level abstract domains can be expressed over regions or arrays but ultimately everything is mapped to ghost (numerical) variables (we call it also base variables) in the base (numerical) domain.
Region variables are never mapped  directly to the base domain. This is simply, because a numerical domain doesn't know what to do with a region. So we need to map a region variable into actually one or more numerical variables. For instance, if the region contains an integer then we have a numerical variable representing that content of the region. If the region contains a reference then the region is mapped to two integers (one per offset and one per size of allocation site).  

So the expand operation (at least in the original region domain) happens at the base (numerical) domain level so there is no regions involved. The `init` and `count` environments map region variables to some piece of information (whether the region is initialized and how many references contains, respectively). The number of region variables is always fixed in the CrabIR program so the abstract domain never creates region variables at runtime.

include/crab/domains/abstract_domain_params.hpp
```diff
@@ -152,6 +152,8 @@ class region_domain_params {
   bool m_is_dereferenceable;
   // reason about unknown regions
   bool m_skip_unknown_regions;
+  // keep track of mru cache
+  bool m_use_mru_regions;
```
caballa: These flags are intended to add extra functionality to the region abstract domain.
Here, we are actually creating a new abstract domain. Thus, we will create a new file called `mru_region_domain.hpp` which can be initially a copy of the original `region_domain.hpp`. Later, we can think of how to factorize code. Both the original `region_domain` and the new `mru_region_domain` can share these flags for now.
But **don't do it now** because otherwise it will be very hard to see which parts are new and which parts are from `region_domain.hpp`.

include/crab/domains/region_domain.hpp
```diff
@@ -2,12 +2,14 @@
 
 #include <crab/domains/abstract_domain.hpp>
 #include <crab/domains/abstract_domain_params.hpp>
+#include <crab/domains/apron_domains.hpp>
```
caballa: Replace with `term_equiv.hpp`. This is the domain that can track equalities between uninterpreted functions.

include/crab/domains/region_domain.hpp
```diff
@@ -561,6 +567,30 @@ class region_domain final
   base_abstract_domain_t m_base_dom;
   /** End base domain **/
 
+  /** Begin MRU domain **/
+  // The cache abstract domain that capture invariants for MRU object
+  base_abstract_domain_t m_cache_dom;
+  using base_variable_domain_t = variable_domain<base_variable_t>;
+  // left: the base address for cache
+  // right: a region variable to get equivalence class
+  using cache_variable_opt_t =
```
caballa: **Note to myself:** REVISIT this: a ghost variable is mapped to region variable.

include/crab/domains/region_domain.hpp
```diff
@@ -756,8 +806,13 @@ class region_domain final
     m_base_dom.rename(left_vars, out_vars);
     right_dom.project(right_vars);
     right_dom.rename(right_vars, out_vars);
+    m_cache_dom.project(left_vars);
+    m_cache_dom.rename(left_vars, out_vars);
+    right_cache_dom.project(right_vars);
+    right_cache_dom.rename(right_vars, out_vars);
 
     m_base_dom |= right_dom;
+    m_cache_dom |= right_cache_dom;
```
caballa: This code is fine because before we join the two caches, we do the proper renaming so that the same program variable of type `variable_t` is mapped to the same variable of type `base_dom_var_t`

include/crab/domains/region_domain.hpp
```diff
@@ -801,9 +858,25 @@ class region_domain final
     // variables).  The domain is finite
     tag_env_t out_tag_env(left.m_tag_env | right.m_tag_env);
 
+    cache_variable_opt_t out_cache_vars = boost::none;
```
caballa: Same problems  with `do_join`

include/crab/domains/region_domain.hpp
```diff
@@ -877,6 +957,20 @@ class region_domain final
                                       right.m_rgn_dealloc_dom);
     tag_env_t out_tag_env(left.m_tag_env & right.m_tag_env);
 
+    cache_variable_opt_t out_cache_vars = boost::none;
```
caballa: same problems with `do_join`

include/crab/domains/region_domain.hpp
```diff
@@ -1325,6 +1428,48 @@ class region_domain final
...
+    if (!m_cache_dom.is_bottom()) {
+        // 1. join cache and base domain
+        m_base_dom |= m_cache_dom;
+        // 1.a update init
+
+        // 1.b update countaddr
+
+        // 2. flush cache
+        if (auto m_cache_rgn_opt = get_cache_rgn_var()) {
+          variable_vector_t mru_rgn_vec =
```
caballa: Don't use `m_rgn_dealloc_dom`. This piece of state is for reasoning about deallocations. Don't mix things.

include/crab/domains/region_domain.hpp
```diff
@@ -561,6 +567,30 @@ class region_domain final
...
+  base_abstract_domain_t m_cache_dom;
+  using base_variable_domain_t = variable_domain<base_variable_t>;
+  // left: the base address for cache
+  // right: a region variable to get equivalence class
+  using cache_variable_opt_t =
+      boost::optional<std::pair<base_variable_domain_t, variable_t>>;
+  cache_variable_opt_t m_cache_vars;
+  // key: a reference variable
+  // value: a base variable in the base domain that represents the base address
+  // of the MRU
```
caballa: Note to myself: this remembers that a reference variable of type `variable_t` is mapped to a ghost variable of type `base_variable_t`

include/crab/domains/region_domain.hpp
```diff
@@ -722,6 +758,20 @@ class region_domain final
 
     base_varname_allocator_t out_alloc(m_alloc, right.m_alloc);
     base_abstract_domain_t right_dom(right.m_base_dom);
+    base_abstract_domain_t right_cache_dom(right.m_cache_dom);
```
caballa: We need to rename `m_base_variable_map` and `right.m_base_variable_map` before join. 
Both maps map from `variable_t` to `base_variable_t`. Let's us consider this code

```
  if (*) {  ref1 := make_ref(rgn1, ...); } else { ref1 := make_ref(rgn1, ...); }
 ```
 In the if branch, the variable `ref1` is mapped to some dynamically created variable `tmp1`.
 Similarly, on the else branch, `ref1` is mapped to a different dynamically created variable `tmp2`.
So, we need to rename `tmp1` and `tmp2` to a common new variable (e.g., `tmp3`) before we join.
This is exactly what we do with `m_base_dom` (the numerical domain) and we need to do also with `m_cache_dom` the cache. Before any lattice operation (join/meet/widening/narrowing/inclusion) we need to do a common renaming.

File: include/crab/domains/region_domain.hpp [hunk 15]
```diff
@@ -2001,6 +2157,14 @@ class region_domain final
       m_alloc_site_dom.set(ref, as);
     }
 
+    /* MRU start */
+    // 1. create a new ref var ref^base
+    base_variable_t ref_base =
+        ghost_variables_t::make_base_variable(m_alloc, ref.get_type());
+    // 2. update it into m_base_variable_map
+    m_base_variable_map.set(ref, ref_base);
```
caballa: comment to my self: why do we need a ghost variable `ref_base`?

include/crab/domains/region_domain.hpp
```diff
@@ -2134,7 +2298,63 @@ class region_domain final
...
+    /* MRU start */
+    if (crab_domain_params_man::get().region_use_mru_regions()) {
+      if (m_cache_vars != boost::none) {
+        bool res = (*m_cache_vars).first == m_base_variable_map[ref];
+        CRAB_LOG("region-mru",
+            CRAB_WARN("region_mru_domain::ref_load: mru updated? ", res););
+      }
+      // check m_base_variable_map for comparing base address
+      if (m_cache_vars == boost::none ||
+          !(*get_cache_ref_var_dom() == m_base_variable_map[ref])) {
```
caballa: This is an important source of confusion. 
This uses syntactic equality between two variables. 
- `get_cache_ref_var_dom()` returns a ghost variable that represents the base address of the memory object modeled by the cache.
- `m_base_variable_map[ref]` returns a ghost variable created when `make_ref` was called.

I've already mentioned it but I don't understand why we use ghost variables here. 
Moreover, we should use a semantic equality by using an abstract domain that can reason about equalities. The reason to use semantic equality is that we can assignments between references. For instance,

```
ref := make_ref(rgn, ...);
(rgn, ref1) := gep_ref(rgn, ref);
(rgn, ref2) := gep_ref(rgn, ref1);
```
Here, `ref`, `ref1`, and `ref2` represents the same thing. If you use the operator `==` then they will be different because they are different syntactically.

include/crab/domains/region_domain.hpp
```diff
@@ -2134,7 +2298,63 @@ class region_domain final
...
+        }
+        commit_cache();
+      }
+    }
+    if (crab_domain_params_man::get().region_use_mru_regions()) {
+      // 3. update cache from base for new MRU object
+      auto rgn_vec = m_rgn_dealloc_dom.get_equiv_vars_by_var(rgn);
+      // copy all relationships between all regions from the new MRU object
+      // FIXME: Invoke general version of expand function
+      // YS: I think we need to update m_rgn_counting_dom for a copied region var
```
caballa: The expand operation operates only on variables of type `base_variable_t`.
So a region variable (which is type `variable_t`) must be first translated to the bunch of ghost variables (of type `base_variable_t`) which model different aspects of the region. Therefore, we don't update `m_rgn_counting_dom` since we never copy region variables

include/crab/domains/region_domain.hpp
```diff
@@ -2134,7 +2298,63 @@ class region_domain final
...
+      ghost_variables_t ref_gvars = get_or_insert_gvars(ref);
+      if (auto region_gvars_opt = get_gvars(rgn)) {
+        ghost_variables_t fresh_region_gvars =
+            ghost_variables_t::create(m_alloc, (*region_gvars_opt));
+        (*region_gvars_opt).expand(m_base_dom, fresh_region_gvars);
```
caballa: I don't know what the expand does here

include/crab/domains/region_domain.hpp
```diff
@@ -2134,7 +2298,63 @@ class region_domain final
...
+        gvars_res.assign(m_cache_dom, fresh_region_gvars);
```
caballa: Comment to myself: this is like a strong read on `m_cache_dom`

include/crab/domains/region_domain.hpp
```diff
@@ -722,6 +758,20 @@ class region_domain final
...
+    cache_variable_opt_t out_cache_vars = boost::none;
+    if (right.m_cache_vars != boost::none) {
+      if (m_cache_vars != boost::none) {
+        out_cache_vars = right.m_cache_vars;
+      }
+      if (m_cache_vars.get().second != right.m_cache_vars.get().second) {
+        CRAB_ERROR(domain_name(), "::do_join_or_widening: mru cache rgn variable should be consistent",
+            m_cache_vars.get().second, "!=", right.m_cache_vars.get().second);
+      }
+      out_cache_vars = cache_variable_opt_t({m_cache_vars.get().first | right.m_cache_vars.get().first, m_cache_vars.get().second});
```
caballa: Note to my self: REVISIT this code. Line 771 is joining ghost variables without renaming which a red flag. 
Moreover, if the two abstract states model different caches then the join shouldn't report an ERROR. Instead, the result of the join should state somehow that there is no cache being tracked.

caballa (commented):
Some general comments:

- Use of `variable_t` VERSUS `base_variable_t` 
  Variables that appear in the CrabIR are of type `variable_t`. Any
  variable created dynamically by the abstract domain are of type
  `base_variable_t`.

  It's very important to keep track which `variable_t` maps to
  which (set of) `base_variable_t`. This is needed because all lattice
  operations (less or equal, join, meet, widening, narrowing) must
  perform some renaming **before** the lattice operation is
  applied. Therefore, if you create a `base_variable_t` and don't rename
  it during the lattice operations then something will go wrong.

- Use of an union-find to group together which regions are originated
  from the same memory object VERSUS keeping track for each region
  which is the program variable (of type `reference`) that represents
  the base address of the memory object and use the `term` domain to propagate equalities so that if two regions are mapped to the "same" variable (where "same" means that the term domain says they are equal) then we know that they are originated from the same memory object. 
  I thought we agreed on the second. But you are reusing some
  union-find used already in the code for something else which is a
  mess. 

- You use ghost variables to remember the variable that represents the
  address returned by a call to `make_ref`. We should use variables of
  type `variable_t`. I don't see a reason to add extra ghost variables unless I'm missing something.

- Syntactic versus semantic equality between variables.
  We need the term domain to keep track of which variables of type
  `variable_t` are equal. Don't use `operator==` from `variable_t`.

- Finally and not less important, it's a mess the cache. This is what I was able to understand. We have two abstract states (using the same domain) that represents the cache (`m_cache_dom`) and what is not the cache (`m_base_dom`).
     - All non-memory operations use `m_base_dom`
     - Memory write use either `m_cache_dom` or `m_base_dom` depending on whether we        
        can use the cache or not.
     - A memory read updates both `m_base_dom` and `m_base_dom`
     
     These patterns are not consistent. What I have in mind was to start using `m_base_dom` for all operations.  Then, as soon as we populate a cache then *all* operations use `m_cache_dom`. The tricky part is when we invalidate a cache. This operation is complex because it needs to project any relationship that is not affected by the memory operations modeled by the cache and meet them with `m_base_dom`. For the memory operations modeled by the cache we need to join them (to perform the weak update). So the invalidation requires probably a couple of calls to project, one call to meet and one call to join. Maybe it can be done better but I think this is conceptually much simpler.

PR #43, opened 2021-12-15
Prototype mru

LinerSu:
1. Prototype a simple update cache if the cache is empty.
2. Prototype basic operations:
    - `fold`: add three examples;
    - `expand`: add two examples.


LinerSu:
Details about my implementation, I made `mru_region_domain` class as a friend class of `region_domain` (also forward declare as well) because all the fields in the original implementation are private members. I think the friend class is the easier way to design so far.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+  using typename abstract_domain_t::reference_constraint_t;
+  using typename abstract_domain_t::variable_or_constant_t;
+  using typename abstract_domain_t::variable_or_constant_vector_t;
+  using typename abstract_domain_t::variable_t;
+  using typename abstract_domain_t::variable_vector_t;
+  using number_t = typename Params::number_t;
+  using varname_t = typename Params::varname_t;
+
+private:
+  using base_uf_domain_t = uf_domain<number_t, varname_t>;
```
caballa: it should use `Params::base_varname_t` instead of `varname_t`

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+  // To create ghost variables for the base domain.
+  base_varname_allocator_t m_alloc;
+  // Map a variable_t to its ghost variables:
+  var_map_t m_var_map;
+  // Reverse map from ghost variables to variable_t
+  rev_var_map_t m_rev_var_map;
+
+public: // FIXME:
+  // The base abstract domains (over ghost variables): the domains only
+  // contain variables
```
caballa: We need `m_mem` of `base_abstract_domain_t` type to represent memory state without cache. @yusen: this is the field https://github.com/seahorn/crab/blob/master/include/crab/domains/region_domain.hpp#L145

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+  base_abstract_domain_t m_cache_lines;
+  base_uf_domain_t m_regs;
```
caballa, line 155: Add comment that says that "it models equalities between scalar variables in `m_mem` and variables in `m_cache_lines`

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+  base_uf_domain_t m_addrs;
```
caballa, line 162: Add comment that says that "two variables are mapped to the same uninterpreted symbol if they belong to the same memory object"

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+                    base_bool_domain_t &&flags)
+      : m_is_bottom(cache_lines.is_bottom()), m_alloc(std::move(alloc)),
+        m_var_map(std::move(var_map)), m_rev_var_map(std::move(rev_var_map)),
+        m_cache_lines(std::move(cache_lines)), m_regs(std::move(regs)),
+        m_addrs(std::move(addrs)), m_flags(std::move(flags)) {}
+
+  // Perform *this = join(*this, right)
+  void do_join(const mru_region_domain_t &right) {
+    base_varname_allocator_t out_alloc(m_alloc, right.m_alloc);
+    base_uf_domain_t out_addrs(m_addrs | right.m_addrs);
```
caballa: `m_addrs` and `m_flags` contain base (ghost) variables so they must be renamed.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+  }
+
+  mru_region_domain_t do_join_or_widening(
+      const mru_region_domain_t &left, const mru_region_domain_t &right,
+      const bool is_join /*unused*/, base_dom_binop_t base_dom_op,
+      base_uf_dom_binop_t base_uf_dom_op) const {
+
+    base_varname_allocator_t out_alloc(left.m_alloc, right.m_alloc);
+    // addrs does not require common renaming (i.e., no ghost variables).
+    base_uf_domain_t out_addrs(m_addrs | right.m_addrs);
```
caballa: Same renaming problem.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+    return res;
+  }
+
+  mru_region_domain_t do_meet_or_narrowing(
+      const mru_region_domain_t &left, const mru_region_domain_t &right,
+      const bool is_meet /*unused*/, base_dom_binop_t base_dom_op,
+      base_uf_dom_binop_t base_uf_dom_op) const {
+
+    // addrs does not require common renaming (i.e., no ghost variables).
+    base_uf_domain_t out_addrs(m_addrs & right.m_addrs);
```
caballa: Same renaming problem

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+                                            << "\n\t" << o << "\n";);
+    if (is_bottom() || o.is_top()) { // this indeed <= o
+      CRAB_LOG("mru-region-leq", crab::outs() << "Result1=1\n";);
+      return true;
+    } else if (is_top() || o.is_bottom()) { // this indeed > o
+      CRAB_LOG("mru-region-leq", crab::outs() << "Result2=0\n";);
+      return false;
+    }
+
+    if (!(m_addrs <= o.m_addrs)) {
```
caballa: it requires common renaming because they contain ghost variables

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+      CRAB_LOG("mru-region-leq", crab::outs() << "Result3=0\n";);
+      return false;
+    }
+
+    if (!(m_flags <= o.m_flags)) {
```
caballa: same renaming problem

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+    std::swap(m_rev_var_map, out_rev_var_map);
+    std::swap(m_addrs, out_addrs);
+    std::swap(m_flags, out_flags);
+
+    CRAB_LOG("mru-region", crab::outs() << *this << "\n");
+  }
+
+  mru_region_domain_t do_join_or_widening(
+      const mru_region_domain_t &left, const mru_region_domain_t &right,
+      const bool is_join /*unused*/, base_dom_binop_t base_dom_op,
```
caballa, line 425: Be consistent. Either pass a function that performs join or widening on each subdomain or use the bool `is_join`. Right now, you pass the function only for `m_cache_lines` and `m_regs` but not for `m_addrs` , etc. My personal taste is to use `is_join` because it reduces the number of parameters so it's less verbose.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+      m_alloc.clear_renaming_map();
+    }
+  }
+  /**------------- End domain API definitions -------------------**/
+
+  /**------------- Begin MRU API definitions -------------------**/
+  // perform a similiar work for cache.lines.constraint(cache.regs)
+  // This one requires to be renaming because mem base dom used different base
+  // vars
+  mru_region_domain_t cstrs() {
```
caballa: it should be private.

caballa: This seems to be part of a reduction step. We pass all the equalities from `m_regs` and add them to `m_cache_lines`. More comments would be helpful

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+    auto csts = m_regs.to_linear_constraint_system();
+    mru_region_domain_t res(*this);
+    res.m_cache_lines += csts;
+    return res;
+  }
+
+  // get all used Crab IR vars by region dom
+  void vars(variable_vector_t &used_vars) const {
```
caballa: it should be private

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+    used_vars.reserve(m_var_map.size());
+    for (auto &kv : m_var_map) {
+      used_vars.push_back(kv.first);
+    }
+  }
+
+  void region_vars(variable_vector_t &used_vars) const {
```
caballa: it should be private

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,1336 @@
...
+    used_vars.reserve(m_var_map.size());
+    for (auto &kv : m_var_map) {
+      if (kv.first.get_type().is_region()) {
+        used_vars.push_back(kv.first);
+      }
+    }
+  }
+
+  // NOTE: For debug use
```
caballa: these methods for debugging should be private. Remember to add `CRAB_LOG` to print extra stuff.

include/crab/domains/region_domain.hpp
```diff
@@ -101,7 +102,8 @@ class region_domain final
   using base_linear_constraint_system_t =
       typename base_abstract_domain_t::linear_constraint_system_t;  
   using base_varname_allocator_t = typename Params::varname_allocator_t;
-
+  friend class mru_region_domain<Params>;
```
caballa: Don't touch `region_domain.hpp`. I prefer for now to be `mru_region_domain` completely independent from `region_domain` at the expense of duplication. One design choice is that `mru_region_domain` could have `region_domain` as a private field to represent the non-cache state. However, since `mru_region_domain` needs to perform all the renaming stuff due to ghost variables I prefer not to call it.

include/crab/domains/mru_region_domain.hpp
LinerSu: this is what I want to use overloading. I saw we pass a function ptr as an argument here. So I expect we can use the template feature to pass just one function. That is, passing a template function (lambda function) as an argument here to allow each individual subdomain to call their corresponded methods. However, I think the template for lambda function is C++ 20 features. I also cannot pass a template function as an argument here. Any idea to achieve this?

include/crab/domains/mru_region_domain.hpp
LinerSu: Agree. This one follows the old implementation, will change that in the next iteration.

include/crab/domains/mru_region_domain.hpp
LinerSu: Is that possible to pass some flags for `test-bin`? I used `test-bin/fold1`, `test-bin/expand1` for unit testing. I know we can pass them through the `clam.py` or `crab` executable.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  // initialized on the right then we extract the information from the
+  // right and add it to the left. This can happen if the first store
+  // to a region happens inside a loop. At the entry of the loop
+  // nothing is known about the region so the join there would lose
+  // all the information about the region.
+  //
+  // This step is not optimal because it keeps only non-relational
+  // information about regions but no relational information between
+  // regions and other variables.
+  void refine_regions(const variable_vector_t &left_regions,
```
caballa, line 272: Remove this function. It's used only when some flag is enabled but it turned out that it's not useful

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+        m_addrs(std::move(addrs)),
+        m_rgn_counting_dom(std::move(rgn_counting_dom)),
+        m_rgn_init_dom(std::move(rgn_init_dom)),
+        m_rgn_type_dom(std::move(rgn_type_dom)),
+        m_alloc_site_dom(std::move(alloc_site_dom)),
+        m_rgn_dealloc_dom(std::move(rgn_dealloc_dom)),
+        m_tag_env(std::move(tag_env)), m_obj_rgn_map(std::move(obj_rgn_map)),
+        m_rev_obj_rgn_map(std::move(rev_obj_rgn_map)) {}
+
+  bool can_propagate_initialized_regions(
```
caballa, line 243: Remove this function (see next comment below)

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  // an abstract domain that keeps equalities of base address between references
+  // we assume each object is allocated by some address called base address.
+  // each reference refers to an object so each reference has additional info
+  // about base address. This domain tracks all base address alloacted so far.
+  // Thus, two variables are mapped to the same uninterpreted symbol
+  // if they belong to the same memory object.
+  base_uf_domain_t m_addrs;
+
+  variable_t *m_base;
+  variable_t *m_used;
```
caballa, line 165: This should be `boolean_value`

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  variable_t *m_dirty;
```
caballa, line 166: `boolean_value`

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  base_uf_domain_t m_regs;
+  // an abstract domain that keeps equalities of base address between references
+  // we assume each object is allocated by some address called base address.
+  // each reference refers to an object so each reference has additional info
+  // about base address. This domain tracks all base address alloacted so far.
+  // Thus, two variables are mapped to the same uninterpreted symbol
+  // if they belong to the same memory object.
+  base_uf_domain_t m_addrs;
+
+  variable_t *m_base;
```
caballa, line 164: It's not urgent but it would be nice to wrap everything related to the cache into one class: `m_cache`,`m_used`, `m_base` and `m_dirty`.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+    base_uf_domain_t right_addrs(right.m_addrs);
+    var_map_t out_var_map, out_cache_var_map;
+    rev_var_map_t out_rev_var_map;
+    obj_rgn_map_t out_obj_rgn_map;
+    rev_obj_rgn_map_t out_rev_obj_rgn_map;
+    // -- Compute common renamings
+    base_variable_vector_t left_vars, right_vars, out_vars;
+    // reserve vec space by upper bound to avoid reallocations
+    size_t num_renamings = left.m_var_map.size();
+    out_cache_var_map = std::move(left.m_cache_var_map);
```
caballa, line 460: The use of `move` is not right. The left and right operands are passed are const&. This is because you can call the join with operands X and Y and then use X or Y again. If you call `move` here and the caller tries to access X then the program will probably crash.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+    base_variable_vector_t left_vars, right_vars, out_vars;
+    // Only left (right) track variables used in left (right) dom only (not in
+    // common)
+    base_variable_vector_t only_left_vars, only_left_out_vars;
+    base_variable_vector_t only_right_vars, only_right_out_vars;
+    // reserve vec space by upper bound to avoid reallocations
+    size_t left_renamings = left.m_var_map.size();
+    size_t right_renamings = right.m_var_map.size();
+    size_t num_renamings = left_renamings + right_renamings;
+    out_cache_var_map = std::move(left.m_cache_var_map);
```
caballa, line 581: See comment above.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  }
+
+  void vars_in_cache(variable_vector_t &used_vars) const {
+    used_vars.reserve(m_cache_var_map.size());
+    for (auto &kv : m_cache_var_map) {
+      used_vars.push_back(kv.first);
+    }
+  }
+
+  variable_vector_t vec_minus(variable_vector_t left, variable_vector_t right) {
```
caballa, line 1026: If you try to compute the set difference of two vectors then please do something like this
```
variable_vector_t res;
std::sort(left.begin(), left.end());
std::sort(right.begin(), right.end());
std::set_difference(left.begin(), left.end(), right.begin(), right.end(), std::back_inserter(res));
```

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  void region_vars_in_mem(variable_vector_t &used_vars) const {
+    used_vars.reserve(m_var_map.size());
+    for (auto &kv : m_var_map) {
+      if (kv.first.get_type().is_region()) {
+        used_vars.push_back(kv.first);
+      }
+    }
+  }
+
+  void region_vars_in_cache(variable_vector_t &used_vars) const {
```
caballa, line 1010: DUPLICATED

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+
+  // get all used Crab IR vars by region dom
+  void vars(variable_vector_t &used_vars) const {
+    used_vars.reserve(m_var_map.size());
+    for (auto &kv : m_var_map) {
+      used_vars.push_back(kv.first);
+    }
+  }
+
+  void region_vars_in_mem(variable_vector_t &used_vars) const {
```
caballa, line 1001: This is inefficient. You can use `boost::transform_iterator` so that you can iterate e.g., over a map but the caller only iterates over the first or the second elements. This avoids makes memory allocations and copies. Of course, you need to adapt the callers of `region_vars_in_mem`. For instance, `vec_minus`  uses it. So you can adapt `vec_minus` to take a pair of iterators rather than a vector. LMK if this is not clear.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  }
+
+  // TODO: need to implement this on next iteration
+  void invalidate_cache() {}
+
+  // fresh cache as: { cache lines: bot, regs: top, addrs: addrs -= C_base }
+  // for addrs, we keep all equalities of base address for each reference
+  // also need to assign cache used flag C_used as false
+  void make_fresh_cache() {
+    m_cache_lines.set_to_bottom();
```
caballa, line 1072: it should be `set_to_top()`

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+    m_regs.set_to_top();
+    m_addrs.set_to_top();
+    if (m_base) {
+      forget(*m_base);
+    }
+    if (m_used) {
```
caballa, line 1078: I understand now why you define `m_used` as a variable. I think this code is super complicated. Why not just a `boolean_value`?

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+
+    // Step 2.
+    //  Our design is joining cache line with memory.
+    //  We need the projection before join because we only fold regions & regs
+    //  We do not affect other properties in mem.
+    variable_vector_t vars_cache;
+    vars_in_cache(vars_cache); // vars for cache lines + registers
+    mru_region_domain_t A = mru_region_domain_t(*this);
+    A.project(vars_cache); // memory object represented in the cache
+    A.m_mem |= reduced;    // join cache with mem,
```
caballa, line 2238: I think before join we need to rename `Ci` (cache names) into `Ri` (mem names).
Update: I also wrote something in `expand`. Is your intention that cache and mem have the same region variable names?

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+    //    { Ci <= Cj <= Ck; Ci <= r <= Ck; ...  }
+    base_abstract_domain_t reduced = cache_lines_regs_reduce();
+
+    // Step 2.
+    //  Our design is joining cache line with memory.
+    //  We need the projection before join because we only fold regions & regs
+    //  We do not affect other properties in mem.
+    variable_vector_t vars_cache;
+    vars_in_cache(vars_cache); // vars for cache lines + registers
+    mru_region_domain_t A = mru_region_domain_t(*this);
```
caballa, line 2236: minor thing but do `mru_region_domain_t A(*this)`

include/crab/domains/mru_region_domain.hpp
caballa: Also, `vars_cache` seems to contain `Ci` variables so if we project `A` (copy of `m_mem`) on those variables you will probably get top since `A` does not contain `Ci` variables.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  /**------------- Begin MRU API definitions -------------------**/
+
+  // The expand performs a copy of region vars from mem into cache_lines
+  //  i.e. by given a most recently used object o, we copy all regions with o
+  //  in mem into the cache:
+  // For instance, o used V1, V2:
+  //    mem: { 1 <= V1 <= 2; V1 < V2; ... }
+  // The cache line will be:
+  //    cache_lines: { 1 <= V1 <= 2; V1 < V2; }
+  void expand(const variable_t &rgn) {
```
caballa, line 2188: The name `expand` already exists. I would call it differently. I would use a name like `materialize` so we use `fold` and `materialize` like in shape abstract domains.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+                   " should be a region");
+      }
+    };
+
+    if (is_bottom()) {
+      return;
+    }
+
+    if (name == "regions_from_memory_object") {
+      // why pass to the outputs vector?
```
caballa, line 2094: It's not. That line calls the method `intrinsic` from a basic block (https://github.com/seahorn/crab/blob/master/include/crab/cfg/cfg.hpp#L2903), not from the abstract domain. That will be called later. 

```
const statement_t *intrinsic(std::string name,
                               const std::vector<variable_t> &lhs,
                               const std::vector<variable_or_constant_t> &args) {
    return insert(new intrinsic_t(name, lhs, args, this));
  }
 ```

File: include/crab/domains/mru_region_domain.hpp [hunk 32]
```diff
@@ -0,0 +1,2270 @@
...
+      // https://github.com/seahorn/clam/blob/ee2f50723a40377864fee7618b90df92d1413a58/lib/Clam/CfgBuilder.cc#L4501
+      // pass region varaibles into cache_map
+      variable_t representative = outputs[0];
```
caballa, line 2097: so just use `inputs` to access to all region variables that are originated from the same dsa node.

caballa: There is no representative. I just pass all the region variables. Not sure why you need a representative.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+      CRAB_ERROR("mru_region_domain::expand the mru object should be found");
+    }
+    Vs_for_new_obj = it_rev_vars->second;
+
+    // expand does not need to perform base variable renaming
+    mru_region_domain_t tmp = mru_region_domain_t(*this);
+    tmp.project(Vs_for_new_obj);
+
+    // perform expand
+    m_cache_lines |= tmp.m_mem;
```
caballa, line 2208: It should be a meet. The join happens in `fold` to simulate weak update.

caballa: We are not renaming the region variables. Maybe it's not needed in the end and this is what you intend.

include/crab/domains/mru_region_domain.hpp
caballa: You can use here the `union_find_domain`. This is like an union-find data structure but with lattice operations. Here, you take the first region variable and call `make` and then with the rest you call `join` (merge). This is how you would put all the region variables in the same equivalence class. 

Having said this, since we only have this kind of intrinsic at the beginning of each function joins shouldn't be a problem because both operands of the join should have the same equivalence classes since they are created at the beginning. So you can also have a vector of sets or something similar and in the binary lattice operations just report an error if both operands do not agree on the same equivalence classes.

include/crab/domains/mru_region_domain.hpp
```diff
@@ -0,0 +1,2270 @@
...
+  // The expand performs a copy of region vars from mem into cache_lines
+  //  i.e. by given a most recently used object o, we copy all regions with o
+  //  in mem into the cache:
+  // For instance, o used V1, V2:
+  //    mem: { 1 <= V1 <= 2; V1 < V2; ... }
+  // The cache line will be:
+  //    cache_lines: { 1 <= V1 <= 2; V1 < V2; }
+  void expand(const variable_t &rgn) {
+    variable_vector_t Vs_for_new_obj;
+    auto it_var = m_rev_obj_rgn_map.find(rgn);
```
caballa, line 2190: see my discussion about `union_find_domain` in the `intrinsic` operation.

LinerSu:
new implementation on PR #60.

PR #44, opened 2022-02-14
New Object domain

LinerSu:
Introduce a new object domain based on DSA node and cache operation.


include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    // 2.2 join the objects' domain over odi map
+    // In the real join operation,
+    // the following join operates on two odi maps containg the same keys
+    // even if it can performed without such assumption.
+    odi_map_t out_odi_map(m_odi_map | right.m_odi_map);
+
+    // The following map update is not necessary because m_flds_rep_map
+    // is fixed after intrinsic calls.
+    // Here, in case to keep operation sound.
+    flds_rep_map_t out_flds_rep_map;
```
agurfinkel: this is wrong. `flds_rep_map` must be a shared map between all values, not a copy!

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+                                       : left.m_base_dom || right.m_base_dom);
+
+    // 2.2 join / widening the objects' domain over odi map
+    // In the real operation,
+    // the following operates on two odi maps containg the same keys
+    // even if it can performed without such assumption.
+    odi_map_t out_odi_map(is_join ? left.m_odi_map | right.m_odi_map
+                                  : left.m_odi_map || right.m_odi_map);
+
+    // The following map update is not necessary because m_flds_rep_map
```
agurfinkel: wrong, see comment above.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    // In the real operation,
+    // the following operates on two odi maps containg the same keys
+    // even if it can performed without such assumption.
+    odi_map_t out_odi_map(is_meet ? left.m_odi_map & right.m_odi_map
+                                  : left.m_odi_map && right.m_odi_map);
+
+    // The following map update is not necessary because m_flds_rep_map
+    // is fixed after intrinsic calls.
+    // Here, in case to keep operation sound.
+    flds_rep_map_t out_flds_rep_map(left.m_flds_rep_map);
```
agurfinkel: remove

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+          "::get_object_rep_by_rgn, the odi map does not include the region ",
+          rgn, " belonging to an abstract object");
+    }
+    return it->second;
+  }
+
+  void get_rgn_vec(variable_t rep, variable_vector_t &rgn_vec) {
+    for (auto kv : m_flds_rep_map) {
+      if (kv.second == rep) {
+        rgn_vec.push_back(kv.second);
```
agurfinkel: `kv.first`? 

Currently, the function fills `rgn_vec` with `rep`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    }
+  }
+
+  static void ERROR_IF_NOT_INT(const variable_t &v, unsigned line) {
+    if (!v.get_type().is_integer()) {
+      CRAB_ERROR(v, ":", v.get_type(), " is not an integer at line ", line);
+    }
+  }
+
+  variable_t get_object_rep_by_rgn(variable_t rgn) {
```
agurfinkel: rename, `get_oid`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+  void get_rgn_vec(variable_t rep, variable_vector_t &rgn_vec) {
+    for (auto kv : m_flds_rep_map) {
+      if (kv.second == rep) {
+        rgn_vec.push_back(kv.second);
+      }
+    }
+  }
+
+  template <class DOM>
+  void assign_interval(DOM &dom, const variable_t v, interval_t ival) {
```
agurfinkel: this is a helper function. It should not be part of the class.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    if (ub) {
+      dom += (v <= *ub);
+    }
+  }
+
+  void move_singleton_to_odi_map(variable_t rep) {
+    variable_vector_t rgn_vec;
+    base_abstract_domain_t tmp_base(m_base_dom);
+    tmp_base.project(rgn_vec);
+    // Be careful of the following operation if base dom and object dom are
```
agurfinkel: you can use a `static_assert` to check for what the comment is saying, or implement a cast operation

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    }
+    if (ub) {
+      dom += (v <= *ub);
+    }
+  }
+
+  void move_singleton_to_odi_map(variable_t rep) {
+    variable_vector_t rgn_vec;
+    base_abstract_domain_t tmp_base(m_base_dom);
+    tmp_base.project(rgn_vec);
```
agurfinkel: `rgn_vec` is empty at this point. Probably forgot to initialize it.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+      if (!x.get_type().is_region()) {
+        // the lhs should be a region variable
+        CRAB_ERROR(name, "::region_assign parameter ", x,
+                   " should be a region");
+      }
+    };
+    error_if_not_rgn(x);
+
+    if (!is_bottom()) {
+      auto it = m_flds_rep_map.find(x);
```
agurfinkel: this needs to be cleaned up. there are functions for this functionality

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+  void assign(const variable_t &x, const linear_expression_t &e) override {
+    crab::CrabStats::count(domain_name() + ".count.assign");
+    crab::ScopedCrabStats __st__(domain_name() + ".assign");
+
+    if (!is_bottom()) {
+      m_base_dom.assign(x, e);
+    }
+  }
+
+  void region_assign(const variable_t &x, const linear_expression_t &e) {
```
agurfinkel: not sure how this function is supposed to be used. 

On one hand, `variable_t &x` is over crab-variables, on the other `linear_expression_t` must be over internal variables of the `odi` representing the object containing `x`. So the client must know both about external and internal variables names to use this function. This cannot be right.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    crab::CrabStats::count(domain_name() + ".count.add_constraints");
+    crab::ScopedCrabStats __st__(domain_name() + ".add_constraints");
+
+    if (!is_bottom()) {
+      m_base_dom += csts;
+    }
+  }
+
+  // add constraint into object domain
+  void add_cons_for_objects(const linear_constraint_t &cst) {
```
agurfinkel: this function is a mess. better remove it before we have a good use for it.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+      m_base_dom -= res;
+      return;
+    }
+
+    variable_t rep = get_object_rep_by_rgn(rgn);
+    // retrieve an abstract object info
+    auto obj_info = m_obj_info_env[rep];
+
+    const small_range &num_refs = obj_info.count_dom();
+    if (num_refs.is_zero()) {
```
agurfinkel: how is this possible? Reading from an object with 0 instances?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+      m_base_dom -= res;
+    } else if (num_refs.is_one()) { // singleton object
+      // strong read
+      m_base_dom.assign(res, rgn);
```
agurfinkel: this should crash. `rgn` is not a variable known to `m_base_dom`. In the worst case, `rgn` is mapped to an index that is known to `m_base_dom` but represents a different variable. Then this will not crash but will produce very strange results.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    } else {
+      // read from odi map
+      // weak read
+      auto obj_dom = m_odi_map[rep];
```
agurfinkel: this is better.

Here, and everywhere, you should be very careful with copy-assignments. I believe that 

```
auto obj_dom = ...
```
is very expensive and should be
```
auto &obj_dom = ...
```

However, you sometimes re-assign `obj_dom`, so you can't use a reference all the time, but should only copy when necessary.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+      return;
+    }
+
+    variable_t rep = get_object_rep_by_rgn(rgn);
+    // retrieve an abstract object info
+    auto obj_info = m_obj_info_env[rep];
+
+    const small_range &num_refs = obj_info.count_dom();
+    if (num_refs.is_zero()) {
+      m_base_dom -= rgn;
```
agurfinkel: I don't think this possible or meaningful

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1303 @@
...
+    } else if (num_refs.is_one()) { // singleton object, strong update
+      if (val.is_constant())
+        m_base_dom.assign(rgn, val.get_constant());
+      else
+        m_base_dom.assign(rgn, val.get_variable());
```
agurfinkel: this is wrong. see comment above

agurfinkel (changes requested):
see comments

include/crab/domains/abstract_domain.hpp
```diff
@@ -414,6 +414,11 @@ template <class Dom> class abstract_domain_api:
     dom.write(o);
     return o;
   }
+
+  friend bool operator==(const Dom& a1, const Dom& a2) {
```
agurfinkel: wrong place. Better to refactor `patricia_tree`. I've sent request to @caballa

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1286 @@
...
+    auto it = (*m_flds_rep_map).find(rgn);
+    if (it == (*m_flds_rep_map).end()) {
+      CRAB_ERROR(domain_name(),
+                 "::get_odi, the odi map does not include the region ", rgn,
+                 " belonging to an abstract object");
+    }
+    return it->second;
+  }
+
+  void get_rgn_vec(variable_t rep, variable_vector_t &rgn_vec) {
```
agurfinkel: rename `get_obj_flds`, flag as potential efficiency issue

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1286 @@
...
+
+    ERROR_IF_NOT_REGION(rgn, __LINE__);
+    ERROR_IF_NOT_REF(ref, __LINE__);
+
+    if (is_bottom()) {
+      return;
+    }
+
+    // The odi must exist for each region after instrisic calls
+    variable_t rep = get_odi(rgn);
```
agurfinkel: I think I missed a function name problem. There are two separate functions:

`get_odi` -- returns an object descriptor given object id
`get_oid` -- returns an object id, given a field.

This should be `get_oid`, and `rep` should be called `oid`.

It would be better to create a new type `oid_t` for object id

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1286 @@
...
+      // set_to_bottom();
+      crab::CrabStats::count(domain_name() +
+                             ".count.ref_load.skipped.null_reference");
+      m_base_dom -= res;
+      return;
+    }
+    // At this point, the requirement for 2 is satisfied
+
+    // The odi must exist for each region after instrisic calls
+    variable_t rep = get_odi(rgn);
```
agurfinkel: `get_oid`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1286 @@
...
+    // So zero case should not exist
+    assert(!num_refs.is_zero());
+
+    // read from odi map
+    const field_abstract_domain_t *obj_dom_ref = m_odi_map.find(rep);
+    if (!obj_dom_ref) {  // not found in odi map, means object domain is top
+      m_base_dom -= res; // forget res means res == top
+      // the requirment 3.1 is satisfied
+    } else {
+      field_abstract_domain_t *obj_dom_ref_no_const =
```
agurfinkel: @caballa please suggest const-safe way to get an interval

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1286 @@
...
+    auto obj_info = m_obj_info_env.at(rep);
+
+    const small_range &num_refs = obj_info.refcount_val();
+    assert(!num_refs.is_zero());
+
+    // read from odi map
+    auto obj_dom_ptr = m_odi_map.find(rep);
+    field_abstract_domain_t *res_obj_dom_ptr;
+    if (obj_dom_ptr) { // found in odi map, create new one by copying old one
+      res_obj_dom_ptr = new field_abstract_domain_t((*obj_dom_ptr));
```
agurfinkel: never ever ever allocate anything on the heap, if you ever need to use `new`, ask.

agurfinkel: the rest of the code is ok, just rewrite the code so that the allocation is on the stack

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1286 @@
...
+  void select(const variable_t &lhs, const linear_constraint_t &cond,
+              const linear_expression_t &e1,
+              const linear_expression_t &e2) override {
+
+    if (!is_bottom()) {
+      m_base_dom.select(lhs, cond, e1, e2);
+    }
+  }
+
+  /********************** Boolean operations **********************/
```
agurfinkel, line 2471: stopped at this point

agurfinkel (changes requested):
some memory related issues. the rest is good. 

does this run on our benchmarks? How does it compare with vstte?

include/crab/domains/abstract_domain.hpp
caballa: this is equality of an abstract domain. what do you want to refactor in `patricia_tree`?

agurfinkel: Ignore this equality. It is wrong. Forget about it.

Next thread. Patricia tree data structure requires various properties of `Value` type. In particular, it assumes that `operator==()` for `Value` is defined. It should not. This is not good C++ design. Instead, it should be parameterized by another parameter `EqualsTo = std::equals_to<Value>`

If it depends on other standard properties of `Value`, they should be exposed in the same way. 
If you want, I can do that.

include/crab/domains/object_domain.hpp
caballa: There are two issues with returning an interval. First, the underlying domain might not keep any interval representation. That's why the returned interval is by value. Second,  the method `interval_t operator[](...)` is not marked as `const` because some domains (e.g., octagons) needs to apply normalization before producing the interval. I could have a second `const` version that cannot normalize.

agurfinkel: A C++ convention is to provide a method called `at()` that is same as `operator[]` but const.

The domain should not expose its internal normalization through `const`. It should hide it using `mutable`. 
At the very least, you want to provide a const `at` method, that, for example, does `const cast` under the hood.

include/crab/domains/abstract_domain.hpp
caballa: OK, I saw now your comments in slack

LinerSu:
@agurfinkel About how object domain compared with VSTTE, I walked through this example.
```
entry:
  region_init(V_i:region(int));
  region_init(V_x:region(int));
  region_init(V_y:region(int));
  i := make_ref(V_i:region(int),4,as_0);
  x := make_ref(V_x:region(int),4,as_1);
  y := make_ref(V_y:region(int),4,as_2);
  store_to_ref(V_i:region(int),i:ref,0:int32);
  store_to_ref(V_x:region(int),x:ref,1:int32);
  store_to_ref(V_y:region(int),y:ref,0:int32);
  goto bb1;
bb1:
  goto bb1_t,bb1_f;
bb1_t:
  *i:int32 := load_from_ref(V_i:region(int),i:ref);
  assume(*i <= 99);
  goto bb2;
bb2:
  *x:int32 := load_from_ref(V_x:region(int),x:ref);
  *y:int32 := load_from_ref(V_y:region(int),y:ref);
  *x = *x+*y;
  store_to_ref(V_x:region(int),x:ref,*x:int32);
  *y:int32 := load_from_ref(V_y:region(int),y:ref);
  *y = *y+1;
  store_to_ref(V_y:region(int),y:ref,*y:int32);
  *i:int32 := load_from_ref(V_i:region(int),i:ref);
  *i = *i+1;
  store_to_ref(V_i:region(int),i:ref,*i:int32);
  goto bb2_end;
bb2_end:
  goto bb1;
bb1_f:
  *i:int32 := load_from_ref(V_i:region(int),i:ref);
  assume(-*i <= -100);
  goto bb3;
bb3:
  *x:int32 := load_from_ref(V_x:region(int),x:ref);
  *y:int32 := load_from_ref(V_y:region(int),y:ref);
  goto ret;
ret:
  assert(-*x+*y <= 0);
```
I use splitDBM for region domain and object domain, the VSTTE can prove assertion in `ret` block. The object domain does not.
The invariants for object domain are:
```
entry={}
bb1=Base = {i-objA<=0, x-objA<=4, y-objA<=8, objA-i<=0, x-i<=4, y-i<=8, objA-x<=-4, i-x<=-4, y-x<=4, x-y<=-4, objA-y<=-8, i-y<=-8},
Object( V_i V_x V_y ) = {V_i -> [0, +oo], V_x -> [1, +oo], V_y -> [0, +oo]}

bb1_f=Base = {i-objA<=0, x-objA<=4, y-objA<=8, objA-i<=0, x-i<=4, y-i<=8, objA-x<=-4, i-x<=-4, y-x<=4, x-y<=-4, objA-y<=-8, i-y<=-8},
Object( V_i V_x V_y ) = {V_i -> [0, +oo], V_x -> [1, +oo], V_y -> [0, +oo]}

bb3=Base = {*i -> [100, +oo], i-objA<=0, x-objA<=4, y-objA<=8, objA-i<=0, x-i<=4, y-i<=8, objA-x<=-4, i-x<=-4, y-x<=4, x-y<=-4, objA-y<=-8, i-y<=-8},
Object( V_i V_x V_y ) = {V_i -> [0, +oo], V_x -> [1, +oo], V_y -> [0, +oo]}

ret=Base = {*i -> [100, +oo], *x -> [1, +oo], *y -> [0, +oo], i-objA<=0, x-objA<=4, y-objA<=8, objA-i<=0, x-i<=4, y-i<=8, objA-x<=-4, i-x<=-4, y-x<=4, x-y<=-4, objA-y<=-8, i-y<=-8},
Object( V_i V_x V_y ) = {V_i -> [0, +oo], V_x -> [1, +oo], V_y -> [0, +oo]}

bb1_t=Base = {i-objA<=0, x-objA<=4, y-objA<=8, objA-i<=0, x-i<=4, y-i<=8, objA-x<=-4, i-x<=-4, y-x<=4, x-y<=-4, objA-y<=-8, i-y<=-8},
Object( V_i V_x V_y ) = {V_i -> [0, +oo], V_x -> [1, +oo], V_y -> [0, +oo]}

bb2=Base = {*i -> [0, 99], i-objA<=0, x-objA<=4, y-objA<=8, objA-i<=0, x-i<=4, y-i<=8, objA-x<=-4, i-x<=-4, y-x<=4, x-y<=-4, objA-y<=-8, i-y<=-8},
Object( V_i V_x V_y ) = {V_i -> [0, +oo], V_x -> [1, +oo], V_y -> [0, +oo]}
```
Here, we lost `V_i == V_x` in fields domain or `*i == *x` in base domain. One reason I think is we lost `V_i == *i` and `V_x == *x` because the `load_ref` or `store_ref` is assigning intervals instead of keeping equality in a domain.

agurfinkel:
@LinerSu looks cool. At this point, I only care about performance. Precision is expected to be very low.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1723 @@
...
+    // Handle case 3, see comments on self_join method.
+    for (auto it = left.m_obj_info_env.begin(); it != left.m_obj_info_env.end();
+         it++) {
+      obj_id_t rep = it->first;
+      const small_range &left_num_refs = it->second.refcount_val();
+      const small_range &right_num_refs =
+          right.m_obj_info_env.at(rep).refcount_val();
+      if (left_num_refs.is_one() &&
+          right_num_refs == small_range::oneOrMore()) {
+        join_singleton_with_non_singleton(rep, left.m_base_dom, out_odi_map);
```
caballa: you need to apply widening if `is_join` is false

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1723 @@
...
+      } else if (right_num_refs.is_one() &&
+                 left_num_refs == small_range::oneOrMore()) {
+        join_singleton_with_non_singleton(rep, right.m_base_dom, out_odi_map);
```
caballa: same here

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1723 @@
...
+    for (auto it = left.m_obj_info_env.begin(); it != left.m_obj_info_env.end();
+         it++) {
+      obj_id_t rep = it->first;
+      const small_range &left_num_refs = it->second.refcount_val();
+      const small_range &right_num_refs =
+          right.m_obj_info_env.at(rep).refcount_val();
+      if (left_num_refs.is_one() &&
+          right_num_refs == small_range::oneOrMore()) {
+        meet_non_singleton_with_singleton(rep, out_base_dom, right.m_odi_map);
+        field_abstract_domain_t out_obj_dom; // top
```
caballa: Here it seems that we are losing a lot of precision by assigning top. why? 
meet should be able to refine things and improve precision.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1723 @@
...
+    if (is_unknown_region(rgn)) {
+      return;
+    }
+
+    if (auto id_opt = get_obj_id(rgn)) {
+      // After instrisic calls, the object id must exist for each region if the
+      // region belongs to an abstract object
+
+      // retrieve an abstract object info
+      auto obj_info = m_obj_info_env.at(*id_opt);
```
caballa: As a general comment, note that `separate_domain` already provides `const Value* find(const Key &k)`. So in cases like this you can avoid copies

include/crab/domains/object_domain.hpp
caballa: Ah, I understood now. Please do `out_odi_map -= rep` What you have it's equivalent but my suggestion states more clear what is going on

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1723 @@
...
+                                         const base_abstract_domain_t &base_dom,
+                                         odi_map_t &odi_map) const {
+    variable_vector_t flds_vec;
+    get_obj_flds(rep, flds_vec);
+
+    base_abstract_domain_t tmp_base(base_dom);
+    tmp_base.project(flds_vec);
+
+    // Be careful of the following operation if type of base dom and object dom
+    // are different
```
caballa: This is something we can check with `static_assert` at the beginning of the class, right?

LinerSu: I used to have it in a function:
https://github.com/LinerSu/crab/blob/5018d40d7b7cca55ad05225e0d2af848f3d42125/include/crab/domains/object_domain.hpp#L547-L552
Now I see the `static_assert` can be a member declaration. I will change that.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,1811 @@
...
+
+    if (!(left.m_obj_info_env <= right.m_obj_info_env)) {
+      CRAB_LOG("object", crab::outs() << "Result3=0\n";);
+      return false;
+    }
+
+    bool res = left.m_base_dom <= right.m_base_dom;
+    CRAB_LOG("object", crab::outs() << "Result4=" << res << "\n";);
+
+    res = left.m_odi_map <= right.m_odi_map;
```
igcontreras: this should be `res = res && (left.m_odi_map <= right.m_odi_map);`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,2378 @@
...
+    // 2. join the odi map
+    base_abstract_domain_t &l_base_dom = m_base_dom;
+    base_abstract_domain_t r_base_dom = right.m_base_dom;
+    odi_map_t out_odi_map;
+    for (auto kv : m_obj_info_env) {
+      const obj_id_t &id = kv.first;
+      const object_info_t &l_obj_info = kv.second;
+      const small_range &l_num_refs = l_obj_info.refcount_val();
+      auto r_obj_info_ref = right.m_obj_info_env.find(id);
+      if (!r_obj_info_ref) {
```
igcontreras, line 363: If an object is not in `right` does it represent that for that object we have top? should the object be forgotten in `this`?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,2378 @@
...
+      : m_is_bottom(base_dom.is_bottom()), m_base_dom(base_dom),
+        m_odi_map(odi_map), m_addrs_dom(addrs_dom), m_uf_regs_dom(uf_regs_dom),
+        m_obj_info_env(obj_info_env), m_flds_id_map(flds_id_map),
+        m_refs_base_addrs_map(refs_base_addrs_map) {}
+
+  // Self join
+  // Perform *this = join(*this, right)
+  void self_join(const object_domain_t &right) {
+
+    // The join operation is sound if:
```
igcontreras: what does 'if' here mean? Are the invariants assumptions of `this` and `right`?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,2378 @@
...
+  // check whether mru object in the left state refers the same as the mru
+  // object in the right state
+  bool
+  mru_refer_same_object(const obj_id_t &id,
+                        const address_abstract_domain_t &l_addrs_dom,
+                        const address_abstract_domain_t &r_addrs_dom) const {
+    variable_t mru_base = get_base_addr_or_fail(id);
+    const usymb_t *l_term_ptr = l_addrs_dom.get_term(mru_base);
+    const usymb_t *r_term_ptr = r_addrs_dom.get_term(mru_base);
+    if (l_term_ptr && r_term_ptr) {
```
igcontreras: Is this code to check if the 2 uninterpreted constants are equal? if `not(a < b)` and `not(b < a)` it could be that `a = b` but it could also be that they are not comparable.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,2378 @@
...
+      bool res = ((*l_term_ptr) < (*r_term_ptr)) == false;
+      res &= ((*r_term_ptr) < (*l_term_ptr)) == false;
+      return res;
+    }
+    return false;
+  }
+
+  bool test_two_addrs_equality(const variable_t &x, const variable_t &y) {
+    const usymb_t *x_term_ptr =
```
igcontreras: same question as in `mru_refer_same_object`. Part of this function could be reused in `mru_refer_same_object`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,2378 @@
...
+    singleton_base.project(flds_vec);
+
+    odi_domain_product_t res_prod; // initial value is top
+    const odi_domain_product_t *prod_ref = s_non_single.m_odi_map.find(id);
+
+    const object_info_t *obj_info_ref = s_non_single.m_obj_info_env.find(id);
+    assert(obj_info_ref);
+    // Be careful of the following operation if type of base dom and object
+    // dom are different
+    if (prod_ref) {
```
igcontreras: This `if` can be moved to line 958

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,2378 @@
...
+      base_abstract_domain_t &non_single_base, const bool is_meet) const {
+
+    const odi_domain_product_t *prod_ref = s_non_single.m_odi_map.find(id);
+
+    const object_info_t *obj_info_ref = s_non_single.m_obj_info_env.find(id);
+    assert(obj_info_ref);
+
+    // Be careful of the following operation if type of base dom and object dom
+    // are different
+    if (prod_ref) {
```
igcontreras: This `if` can be moved to line 986

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,2378 @@
...
+      } else { // num_refs > 1, use odi map
+        // read from odi map
+        const odi_domain_product_t *prod_ref = m_odi_map.find((*id_opt));
+        if (prod_ref != nullptr) {
+          term::term_operator_t symb =
+              invalidate_cache_if_miss((*id_opt), ref, rgn);
+          if (res.get_type().is_reference()) {
+            // res is a reference, assign a fresh symbol into address dom
+            // create a fresh symbol to represent its base address
+            m_addrs_dom.set(get_or_insert_base_addr(res),
```
igcontreras, line 1606: Why new symbol? It should be the same symbol as the base address of what is currently loaded in the cache

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+      return;
+    }
+
+    // REDUCTION: perform reduction
+    perform_reduction();
+
+    if (get_obj_id(rgn) == boost::none) {
+      // if a region does not belong to an object, treat it as an object
+      // treat region as a field, as well as an object id
+      update_fields_id_map(rgn, rgn);
```
igcontreras: In these 2 functions you are doing `(*m_flds_id_map).find(rgn)` twice. See if this can be improved

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+    if (auto id_opt = get_obj_id(rgn)) {
+      // the object id must exist for each region if the
+      // region belongs to an abstract object.
+
+      // retrieve an abstract object info
+      auto obj_info_ref = m_obj_info_env.find(*id_opt);
+      if (!obj_info_ref) { // object goes to top
+        return;
+      }
+      auto old_obj_info = *obj_info_ref;
```
igcontreras: Is this a copy? later you are creating a new one to insert using `m_obj_info_env.set`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+  void ref_store(const variable_t &ref, const variable_t &rgn,
+                 const variable_or_constant_t &val) override {
+    crab::CrabStats::count(domain_name() + ".count.ref_store");
+    crab::ScopedCrabStats __st__(domain_name() + ".ref_store");
+
+    // Perform a store.
+    // ref_store(ref, rgn, val);
+    // Definition: assign abstract value in val to rgn
+    //  rgn: a region variable used in object domain
+    //  res: a register variable used in base domain
```
igcontreras: `val`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+    if (auto id_opt = get_obj_id(rgn)) {
+      // the object id must exist for each region if the
+      // region belongs to an abstract object.
+
+      // retrieve an abstract object info
+      auto obj_info_ref = m_obj_info_env.find(*id_opt);
+      if (!obj_info_ref) { // object goes to top
+        return;
+      }
+      assert(obj_info_ref); // The object info must exsits
```
igcontreras: exist

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+                object_domain_impl::get_or_make_term_symbol(
+                    m_uf_regs_dom,
+                    val_gvars.get_offset_and_size().get_offset()),
+                object_domain_impl::get_or_make_term_symbol(
+                    m_uf_regs_dom, val_gvars.get_offset_and_size().get_size()));
+          }
+        }
+        invalidate_cache_if_miss((*id_opt), ref, rgn, reg_symb,
+                                 reg_offset_size_symb);
+        prod_ref = m_odi_map.find((*id_opt));
```
igcontreras, line 1989: duplicated? (see `const odi_domain_product_t *prod_ref = m_odi_map.find((*id_opt));`)

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+      if (ref_cst.is_tautology()) {
+        return;
+      }
+      if (ref_cst.is_contradiction()) {
+        set_to_bottom();
+        return;
+      }
+
+      // REDUCTION: perform reduction
+      perform_reduction();
```
igcontreras: why here?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+
+    crab::CrabStats::count(domain_name() + ".count.ref_to_int");
+    crab::ScopedCrabStats __st__(domain_name() + ".ref_to_int");
+
+    ERROR_IF_NOT_REF(ref_var, __LINE__);
+    ERROR_IF_NOT_INT(int_var, __LINE__);
+
+    if (!is_bottom()) {
+      // REDUCTION: perform reduction
+      perform_reduction();
```
igcontreras: why here?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+      // ref_var is a reference, assign a fresh symbol into address dom
+      // create a fresh symbol to represent its base address
+      m_addrs_dom.set(get_or_insert_base_addr(ref_var),
+                      object_domain_impl::make_fresh_symbol(m_addrs_dom));
+    }
+  }
+
+  // Make a copy of a region
+  // copy from rhs to lhs
+  void region_copy(const variable_t &lhs_rgn,
```
igcontreras, line 2163: TODO?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+  // x := y op z
+  void apply(arith_operation_t op, const variable_t &x, const variable_t &y,
+             const variable_t &z) override {
+
+    crab::CrabStats::count(domain_name() + ".count.apply");
+    crab::ScopedCrabStats __st__(domain_name() + ".apply");
+
+    if (!is_bottom()) {
+      // REDUCTION: perform reduction
+      perform_reduction();
```
igcontreras: If reducing is a problem, add a flag about what has not been reduced.

For example, if we have several operations with registers
```
x++;
y = y + x;
z = 2 * y;
```

right now we reduce after every operation. you can have a flag per object + base
dom of whether they have been modified since the last reduction.

When you reach a reduction point, do not transfer from objects that have not been modified since the last reduction.

Also, if the operations only involve register arithmetic, there is nothing to reduce until there is a store using those registers.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+    for (auto v : obj_ghost_flds) {
+      if (v.has_offset_and_size()) {
+        obj_dom_flds.push_back(v.get_offset_and_size().get_offset());
+        obj_dom_flds.push_back(v.get_offset_and_size().get_size());
+      }
+      obj_dom_flds.push_back(v.get_var());
+    }
+  }
+
+  void update_fields_id_map(const variable_t &rgn, const obj_id_t &id) {
```
igcontreras, line 780: fields -> flds to be consistent?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+  ghost_variables_t get_or_insert_gvars(const variable_t &v) {
+    return m_ghost_var_man.get_or_insert(v);
+  }
+
+  boost::optional<ghost_variables_t> get_gvars(const variable_t &v) const {
+    return m_ghost_var_man.get(v);
+  }
+
+  base_dom_variable_or_constant_t
+  rename_variable_or_constant(const variable_or_constant_t &v) {
```
igcontreras, line 364: is this equivalent to what is being done in the region domain?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+                                  << "odi = ";
+             odi_write(crab::outs(), prod); crab::outs() << "\n";);
+  }
+
+  // commit cache contents into summary if cache is dirty
+  // reset cache:
+  //  (1) perform reduction by transfering regs-flds' constriants between cache
+  //  domain and base domain (2) clear UF field domain (3) clear cache domain
+  void
+  commit_cache_if_dirty(base_abstract_domain_t &base_dom,
```
igcontreras: if the object is top, just reset the cache/do nothing currently there is a lot of renaming/euf domain reasoning that could be fruitless

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3312 @@
...
+        ghost_variables_t new_gvars = get_or_insert_gvars(new_var);
+        new_gvars.add(new_ghost_vars);
+      } else {
+        continue;
+      }
+    }
+    uf_fld_dom.rename(old_ghost_vars, new_ghost_vars);
+  }
+
+  void forget_a_field(const variable_t &v) {
```
igcontreras: `forget_a_field` -> `forget_fld`?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+
+  static_assert(std::is_same<eq_register_abstract_domain_t,
+                             eq_fields_abstract_domain_t>::value,
+                "eq_register_abstract_domain_t and eq_fields_abstract_domain_t "
+                "must be the same");
+
+  /**------------------ Begin type definitions ------------------**/
+  // Object identifier / object id
+  using obj_id_t = variable_t;
+  using obj_id_set_t = std::set<variable_t>;
```
igcontreras: not used elsewhere?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+    get_obj_flds_with_ghost(id, flds_vec);
+
+    const odi_domain_product_t *prod_ref = m_odi_map.find(id);
+    if (!prod_ref) {
+      CRAB_ERROR(domain_name(), "::", __func__, ": object ", id,
+                 " value is not found on odi map");
+    }
+    // NOTE: copy the following value is required because
+    // updating the odi map is construcing a new tree based on previous tree
+    odi_domain_product_t res_prod = *prod_ref;
```
igcontreras, line 886: is this a deep copy?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+    if (symb_ptr) {
+      return *symb_ptr;
+    } else {
+      return symbolic_variable_equiality_domain_impl::make_fresh_var_symbol<
+          symbolic_variable_equiality_domain_impl::symbolic_var>();
+    }
+  }
+
+  /***************** ODI map operations *****************/
+  /// @brief move singleton object into odi map
```
igcontreras: update comment

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+  /// @pre the precondition is the object now becomes non-singleton
+  /// @param id an object id
+  void move_singleton_to_odi_map(const obj_id_t &id) {
```
igcontreras: rename function since now singletons can be in the odi map

include/crab/domains/object_domain.hpp
igcontreras: Is this copy required if singletons are not in base?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+    // update odi map
+    m_odi_map.set(id, res_prod);
+
+    CRAB_LOG("object-move-odi", crab::outs() << "move object: " << id << "\n";
+             m_odi_map.odi_write(crab::outs(), res_prod);
+             crab::outs() << "\n";);
+  }
+
+  /***************** Cache operations *****************/
+  // commit cache contents into summary
```
igcontreras: if the object is singleton, you can do a reassign instead of join. I don't know if this case actually happens, but if executing the flush cache intrinsic causes the cache to be flushed and then the cache is updated you would lose precision here

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+  commit_cache_if_dirty(base_abstract_domain_t &base_dom,
+                        const eq_register_abstract_domain_t &eq_regs_dom,
+                        object_value_t &prod,
+                        const object_domain_impl::object_info &obj_info_ref,
+                        const obj_id_t &id) const {
+    base_dom_variable_vector_t obj_flds;
+    get_obj_flds_with_ghost(id, obj_flds);
+    // (1) perform reduction by transfering regs-flds' constriants between cache
+    //  domain and base domain
+    reduce_regs_flds_between_cache_and_base(base_dom, eq_regs_dom, prod,
```
igcontreras: is this required for soundness?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+      // commit cache if the cache is dirty
+      commit_cache(sum, cache);
+    }
+    // (2) clear cache domain
+    cache.set_to_top();
+    // (3) clear eq field domain
+    eq_fld.set_to_top();
+  }
+
+  /// @brief
```
igcontreras: doc

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+    }
+    // update odi map
+    m_odi_map.set(id, res_prod);
+
+    CRAB_LOG("object-move-odi", crab::outs() << "move object: " << id << "\n";
+             m_odi_map.odi_write(crab::outs(), res_prod);
+             crab::outs() << "\n";);
+  }
+
+  /***************** Cache operations *****************/
```
igcontreras, line 929: Shouldn't this code be part of the object class functionality?
I see that some interaction is required with the base domain, like reduction, but can't you can pass that as a parameter of methods, for example in commit cache?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+                      std::get<1>(*offset_size_symb));
+    }
+
+    // Step5: update odi map
+    m_odi_map.set(id, odi_domain_product_t(out_obj_info, out_prod));
+  }
+
+  /// @brief check whether we need to commit cache before join / meet
+  /// @return true if the state is required
+  bool commit_is_required() const {
```
igcontreras: this checks if **one** cache needs to be committed but then it is used to commit **all** caches, why not do this check when deciding to commit?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+      const object_info_t &obj_info = odi.first();
+      if (obj_info.cachedirty_val().is_true()) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /// @brief commit all caches on the odi map
+  void commit_all_caches() {
```
igcontreras: This is expensive, for reduction from base to cache. How do you deal with octagon completion before projection, is it cached and then projected into the different objects?

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+      if (prod_ref) {
+        const object_info_t &out_prod_info = prod_ref->first();
+        const small_range &num_refs = out_prod_info.refcount_val();
+        if (num_refs.is_one() &&
+            crab_domain_params_man::get()
+                .singletons_in_base()) { // singleton object
+          continue;
+        }
+        object_value_t out_prod_val = prod_ref->second();
+        reduce_flds_from_base_to_cache(m_base_dom, m_eq_regs_dom, out_prod_val,
```
igcontreras: this could repeat completion when projecting the base domain many times.

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+    }
+
+    CRAB_LOG("object-leq", crab::outs() << "Inclusion test:\n\t" << *this
+                                        << "\n\t" << o << "\n";);
+
+    auto leq_lambda = [this](const object_domain_t &left, const object_domain_t &right) -> bool {
+      return less_than_eq(left, right);
+    };
+
+    if (commit_is_required() && o.commit_is_required()) {
```
igcontreras: see my comment for `commit_is_required`

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+    //                eq_flds_dom = { y == t1; w == t2 }
+    //               )
+    reduce_regs_from_cache_to_base(base_dom, eq_regs_dom, prod, obj_flds);
+    reduce_flds_from_base_to_cache(base_dom, eq_regs_dom, prod, obj_flds);
+  }
+
+  /// @brief a method to invoke domain reduction
+  /// @param reduce_for_assert a special bool indicate reduction is called
+  /// before assertion
+  void perform_reduction(bool reduce_for_assert = false) {
```
igcontreras: rename to capture that it depends on flags. When I read the code later I doubt if it is reducing it always

include/crab/domains/object_domain.hpp
```diff
@@ -0,0 +1,3681 @@
...
+    } else {
+      join_lambda(o);
+    }
+  }
+
+  object_domain_t operator|(const object_domain_t &o) const override {
+    crab::CrabStats::count(domain_name() + ".count.join");
+    crab::ScopedCrabStats __st__(domain_name() + ".join");
+
+    // Trivial cases first
```
igcontreras, line 1501: Are the trivial cases duplicated for every domain?

igcontreras (changes requested):
I looked at `object_domain.hpp` I did not finished. I think some of the performance comments will be addressed when we have the reduction flags that we talked about.

I see some comments from my last review are still here. Please either reply to them or resolve them so we can track progress better.

include/crab/domains/object_domain.hpp
LinerSu: This function will be called if we evaluate either DSA intrinsic or region int statement in crabIR.

include/crab/domains/object_domain.hpp
LinerSu: Yes, even if a singleton object is in the base because the ODI map stores a pair of `<object id, <info, value>>` where the info stores cache status and reference counting. If a singleton in base, the `value` is a top. Copying a top value is cheap.

include/crab/domains/object_domain.hpp
LinerSu: No, I added this because we no longer keep equality between fields and registers once the cache is flushed. If we commented this out. We just lost numerical properties for equality (i.e. imprecise but still sound).

igcontreras:
It seems that you are adding new commits by @caballa and others to master, can you update so that we only see the changes of your commits in the review?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+  }
+};
+
+struct symb_var_hash {
+  std::size_t operator()(const symbolic_var &k) const {
+    return std::hash<uint32_t>()(k.var);
+  }
+};
+
+template <class InputIt1, class InputIt2>
```
igcontreras, line 51: document this function, what does it do? why is it needed?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+bool set_is_absolute_complement(InputIt1 first1, InputIt1 last1,
+                                InputIt2 first2, InputIt2 last2) {
```
igcontreras, line 53: use begin/end names if they are iterators

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+  while (first1 != last1 && first2 != last2) {
+    if (*first1 < *first2)
```
igcontreras, line 55: It is assumed that the elements in the set are ordered, this should be documented

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+  return true;
+}
+
+template <class SYMBVAR> SYMBVAR make_fresh_var_symbol() {
+  return SYMBVAR(term::term_op_val_generator_t::get_next_val());
+}
+
+template <class Domain> class equivalence_class {
+private:
+  std::size_t m_size;
```
igcontreras, line 72: do we need to use `size_t` for this field? How big do we expect equivalence classes to be?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    }
+  }
+  return true;
+}
+
+template <class SYMBVAR> SYMBVAR make_fresh_var_symbol() {
+  return SYMBVAR(term::term_op_val_generator_t::get_next_val());
+}
+
+template <class Domain> class equivalence_class {
```
igcontreras, line 70: If `equivalence_class` is not going to be used outside of this file, you can hide it and make it a struct. Then you don't need to make getters and setters

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+  }
+  o << "}";
+}
+} // namespace symbolic_variable_equiality_domain_impl
+
+#define SVEQ_DOMAIN_SCOPED_STATS(NAME) CRAB_DOMAIN_SCOPED_STATS(NAME, 1)
+#define SVEQ_COUNT_STATS(NAME) CRAB_DOMAIN_COUNT_STATS(NAME, 0)
+
+/// @brief An abstract domain represents equalities used in analyses such as
+/// allocation-site abstraction or domain reduction. In short, giving two
```
igcontreras, line 118: fix grammar

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+  //                #var1        #var2          #var3
+  //  m_classes       ▲            ▲              ▲
+  // ─ ─ ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ┼ ─ ─ ─ ─ ─ ─ ─│─ ─ ─ ─ ─ ─ ─ ─
+  //  m_parents      ┌─┐          ┌┴┐            ┌┴┐
+  //             ┌──▶│1│◀─┐       │4│◀─┐         │6│
+  //             │   └─┘  │       └─┘  │         └─┘
+  //            ┌─┐      ┌─┐          ┌─┐
+  //            │2│      │5│          │3│
+  //            └─┘      └─┘          └─┘
+  // Note that, { 6 } |-> #var3 will be removed after normalization since
```
igcontreras, line 174: update this comment, this is not the case anymore, right?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    for (auto &kv : m_parents) {
+      element_t rep = kv.second; // already path compressed
+      element_set_t &s = res[rep];
+      auto it = std::upper_bound(s.begin(), s.end(), kv.first);
+      s.insert(it, kv.first);
+    }
+    return res;
+  }
+
+  element_set_t get_all_members() const {
```
igcontreras, line 217: not used, remove

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    if (it == m_parents.end()) {
+      return boost::none;
+    }
+    return it->second;
+  }
+
+  /// @brief union two sets
+  /// @param x one set in domain
+  /// @param y one set in domain
+  void union_sets(const element_t &x, const element_t &y) {
```
igcontreras, line 523: This doesn't seem to be used but it looks like `union` operation. Use this instead of overwritting values in `m_parents` directly (see the example in `add`). Try to find all the places where you should be using `union` (`meet` probably)

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    const equivalence_class_t &ec_x = m_classes.at(rep_x);
+    return ec_x.get_absval();
+  }
+
+  /// @brief Add y into the equivalence class of x
+  /// @param x an element may exists
+  /// @param y an element may exists
+  /// @note  if x does not exist, nothing changed.
+  ///        if y \in some cls, forget it from cls, add y into x's class
+  void add(const element_t &x, const element_t &y) {
```
igcontreras, line 640: why not call this one assign

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+      // _|_ <= e || this <= top
+      return true;
+    } else if (is_top() || e.is_bottom()) {
+      // top >= e || this >= _|_
+      return false;
+    }
+    // Return true if *this is a refined partitioning of o
+    //    forall cls_x \in this. exists cls_y \in e ::
+    //      cls_y <= cls_x
+    equivalence_class_elems_t left_equiv_classes =
```
igcontreras, line 695: This is not efficient.

The L <= R operation should check that all equalities that exist in R, exist in L as well (but L may have additional equalities). So you do not need to build equivalence classes of L, once you have an equivalence class of R, simply check that these equalities exists in L.

As soon as you find the first equality from R that is not in L, the answer to <= is false, you can return false and do not compute for the rest of the classes.

With this, you can avoid building all equivalence classes for left and right.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    }
+  }
+
+  /// @brief assign bool lhs := rhs, not used
+  /// @param lhs an element
+  /// @param rhs an element
+  /// @param is_not_rhs if true the assignment is lhs := not(rhs)
+  void assign_bool_var(const element_t &lhs, const element_t &rhs,
+                       bool is_not_rhs) {
+    // WARN: this should not happen
```
igcontreras, line 791: Add `SASSERT(false && "this domain does not treat equalities between boolean vars")`

But why shouldn't it happen? Maybe for the object domain it is not expected to happen.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    return *this;
+  }
+
+  this_domain_t operator&&(const this_domain_t &e) const {
+    SVEQ_DOMAIN_SCOPED_STATS(".narrowing");
+    return *this;
+  }
+
+  /// @brief expand what x equals to to y. This is equivalent to add
+  /// @param x the orginal variable in some class
```
igcontreras, line 767: typo

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    }
+    o << "}";
+  }
+
+public:
+  /**------------------ Begin union find APIs ------------------**/
+  /// @brief Check whether domain constains element v in some class
+  /// @param v an element
+  /// @return true if the elment exists; otherwise, false.
+  bool contains(const element_t &v) const {
```
igcontreras, line 477: I would rename this to `has_representative` or `has_definition`

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+      return;
+    }
+    if (!contains(x)) {
+      set(x, symbolic_variable_equiality_domain_impl::make_fresh_var_symbol<
+                 symbolic_variable_equiality_domain_impl::symbolic_var>());
+    }
+    if (contains(y)) {
+      *this -= y;
+    }
+    element_t rep_x = find(x);
```
igcontreras, line 651: This should be done by calling the `union` operation. You are not updating the size of the class

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+      ec_y.get_size() += ec_x.get_size();
+    }
+  }
+
+  /// @brief set a domain value to x's class
+  /// @param x an element
+  /// @param absval a domain value
+  /// @note  if x does not exist, create an class with element x and absval
+  ///        if x \in some cls, update current domain value by absval
+  void set(const element_t &x, domain_t absval) {
```
igcontreras, line 559: `set_class_value` instead?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+      element_t rep_x = find(x);
+      equivalence_class_t &ec_x = m_classes.at(rep_x);
+      ec_x.set_absval(absval_ptr);
+    }
+  }
+
+  /// @brief get the domain value stored in current equivalent class
+  /// @param v an element in some set
+  /// @return null if !contains(x), otherwise, returns a shared pointer to value
+  std::shared_ptr<domain_t> get(const element_t &x) {
```
igcontreras, line 580: `get_class_value`?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+        return get_all_members_from_an_equiv_class(key);
+      }
+    }
+    return boost::none;
+  }
+
+  /// @brief get the domain value stored in current equivalent class
+  /// @param v an element in some set
+  /// @return null if !contains(x), otherwise, returns a shared pointer to value
+  std::shared_ptr<const domain_t> get(const element_t &x) const {
```
igcontreras, line 622: put this together with the other `get`

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+  /// TODO: This operation is not an optimal solution. This computation does
+  /// not follow the feature of union-find data structure.
+  /// The solution is merging two equivalence classes if two equivalence classes
+  /// shared common elements. As the result, it computed a set of a set of
+  /// equivalence classes where each class is a super set of two or more classes
+  /// from the given two abstract states. That is, the result of the meet will
+  /// keep equalities appeared either in left or in right states.
+  /// Formally, forall cls_x \in left. forall cls_y \in right ::
+  ///           cls_x ^ cls_y != \empty => cls_x union cls_y
+  this_domain_t meet(const this_domain_t &left,
```
igcontreras, line 365: Building the classes from left and right is not efficient. Try doing this:

Initialize `res` as a copy of `left` (or `right`). For every equality `a = b` in `right` (or `left`), call `res.union(a,b)`.

What about abstract values of the classes, are you doing the meet for those?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+  /// given two abstract states. That is, the result of the join only remains
+  /// equalities appeared both in left and in right states.
+  /// Formally, forall cls_x \in left. forall cls_y \in right ::
+  ///           cls_x ^ cls_y
+  /// 
+  /// This operation is running on Quadratic Time -- O(n^2).
+  /// The computation follows the structure of unordered_map.
+  /// The operation requires to construct a new unordered_map.
+  /// TODO: to improve sharable results, we need to change data structure.
+  this_domain_t join(const this_domain_t &left,
```
igcontreras, line 300: what about the abstract values of equivalence classes? is join being done?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+                       bool is_not_rhs) {
+    // WARN: this should not happen
+    return;
+  }
+
+  /// @brief remove an element from a class
+  /// @param v an element
+  /// @note  if v \in cls \land v is representative, select a new representative
+  ///        before removing v.
+  void operator-=(const element_t &v) {
```
igcontreras, line 799: Decrease the size of the class of `v`.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    for (auto v : elements) {
+      *this -= v;
+    }
+  }
+
+  /// @brief keep equivalence classes only for element in vector elements
+  /// @param elements a vector of elements required to keep
+  /// @note  After project operation, the state only keeps equivalence classes
+  ///        for element in vector elements
+  void project(const std::vector<element_t> &elements) {
```
igcontreras, line 848: Is this maintaining the sizes of the classes?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+      const element_t &rep = kv.first;
+      const element_set_t &s = kv.second;
+      bool keep_rep = false;
+      for (auto &e : s) {
+        if (e == rep) {
+          keep_rep = true;
+          break;
+        }
+      }
+      element_t new_rep = keep_rep ? rep : s[0];
```
igcontreras, line 880: Why create new representatives? Can't you just forget what is not in `elements`?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+    SVEQ_DOMAIN_SCOPED_STATS(".rename");
+
+    if (is_top() || is_bottom()) {
+      return;
+    }
+    if (old_elements.size() != new_elements.size()) {
+      CRAB_ERROR(domain_name(),
+                 "::rename with input vectors of different sizes");
+    }
+    for (unsigned i = 0, size = old_elements.size(); i < size; ++i) {
```
igcontreras, line 903: use iterators, they are more efficient

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+          it = m_parents.erase(it);
+        } else if ((*it).second == old_v) {
+          // find it as value
+          (*it).second = new_v;
+          ++it;
+        } else {
+          ++it;
+        }
+        if (parent_old_v) {
+          // insert a key-value pair for new_v
```
igcontreras, line 927: is it possible to replace a key instead of remove and insert?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1031 @@
...
+
+    CRAB_LOG("symb-var-eq", crab::outs()
+                                << "Before normalization " << *this << "\n";);
+    // TODO: could implement to check each size of class instead of creating
+    // equivalence_class_elems_t map
+    equivalence_class_elems_t equiv_classes = equiv_classes_elems();
+    for (auto &kv : equiv_classes) {
+      const element_t &rep = kv.first;
+      const element_set_t &s = kv.second;
+      if (s.size() == 1) {
```
igcontreras, line 958: This is not what we talked about. If our equality domains are spread out, the value of the class matters even if the class is of size 1

igcontreras (commented):
First pass over equality domain

include/crab/domains/uf_domain.hpp
igcontreras, line 1: Are you using this domain? Should this be removed?

caballa:

PR #49, opened 2022-08-08
feat(fixed_tvpi): implement partially binary operation with variable operands

LinerSu:
A rewrite rule for binary operation `x = y * z` if `y` or `z`'s value is constant.

### DIFF
--- a/include/crab/domains/fixed_tvpi_domain.hpp
+++ b/include/crab/domains/fixed_tvpi_domain.hpp
@@ -225,8 +225,14 @@ private:
 
     if (boost::optional<number_t> n = is_var_constant(z)) {
       rewrite_apply(op, x, y, *n, coefficient);
-    } else if (boost::optional<number_t> n = is_var_constant(y)) {
-      rewrite_apply(op, x, z, *n, coefficient);
+    } else if (op == OP_ADDITION || OP_MULTIPLICATION) {
+      // ignore case if op is a subtraction or division
+      if (boost::optional<number_t> n = is_var_constant(y)) {
+        rewrite_apply(op, x, z, *n, coefficient);
+      } else {
+        variable_t ghost_x = get_ghost_var(x, coefficient);
+        m_base_absval -= ghost_x;
+      }
     } else {
       variable_t ghost_x = get_ghost_var(x, coefficient);
       m_base_absval -= ghost_x;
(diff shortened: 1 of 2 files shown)


include/crab/domains/fixed_tvpi_domain.hpp
```diff
@@ -213,6 +213,26 @@ class fixed_tvpi_domain
...
+
+  void rewrite_apply_var(arith_operation_t op, const variable_t &x,
+                         const variable_t &y, const variable_t &z,
+                         unsigned coefficient) {
+    assert(coefficient > 1);
+
+    if (boost::optional<number_t> n = is_var_constant(z)) {
+      rewrite_apply(op, x, y, *n, coefficient);
+    } else if (boost::optional<number_t> n = is_var_constant(y)) {
+      rewrite_apply(op, x, z, *n, coefficient);
```
caballa: This assumes that `op` is commutative so it won't work for subtraction and division.

LinerSu: Yes, you are right. If `y` is some constant `n`, then we should have a special rewrite for `x := n - z` or `x := n / z`. How about adding constraints for those two cases?

caballa: Unless we need those cases now I would just exclude them (i.e., do nothing if sub or div)

LinerSu: Ok. Let me add one condition based on this.

LinerSu: DONE.

include/crab/domains/fixed_tvpi_domain.hpp
```diff
@@ -213,6 +213,29 @@ class fixed_tvpi_domain
...
+  void rewrite_apply_var(arith_operation_t op, const variable_t &x,
+                         const variable_t &y, const variable_t &z,
+                         unsigned coefficient) {
+    assert(coefficient > 1);
+
+    if (boost::optional<number_t> n = is_var_constant(z)) {
+      rewrite_apply(op, x, y, *n, coefficient);
+    } else if (op == OP_ADDITION || OP_MULTIPLICATION) {
+      // Ignore case if op is subtraction and division
+      if (boost::optional<number_t> n = is_var_constant(y)) {
```
caballa, line 230: you forgot `m_base_absval -= ghost_x;` in case of sub or div

LinerSu: fixed

PR #52, opened 2022-10-27
PPLite integration using apron

gretadolcetti:
PPLite domains are now integrated inside crab as apron domains.


cmake/FindFLINT.cmake
```diff
@@ -0,0 +1,28 @@
+# Find FLINT headers and libraries.
+# Author: Enea Zaffanella (enea.zaffanella@unipr.it)
+
+if (NOT FLINT_FOUND)
+  set(FLINT_ROOT "" CACHE PATH "Path to flint install directory")
+
+  find_path(FLINT_INCLUDES
+    NAMES flint/flint.h
+    HINTS "${FLINT_ROOT}/flint"
```
caballa: This didn't work for me. It needs to be 

```
HINTS "${FLINT_ROOT}/include"
```

cmake/download_pplite.cmake
```diff
@@ -0,0 +1,13 @@
+if (CRAB_USE_PPLITE)
+  ExternalProject_Add(pplite
+    URL https://github.com/ezaffanella/PPLite/raw/main/releases/pplite-0.7.1.tar.gz
+    INSTALL_DIR ${CMAKE_BINARY_DIR}/run/pplite
+    CONFIGURE_COMMAND
+    ../pplite/configure --prefix <INSTALL_DIR> --enable-apron --with-gmp=${GMP_SEARCH_PATH} --with-flint=${FLINT_SEARCH_PATH}
```
caballa: This assumes apron is found. For me, I needed to add an extra option to `configure` command:
```
--with-apron=${APRON_ROOT}
```

include/crab/domains/apron_domains.hpp
```diff
@@ -107,6 +112,19 @@ class apron_domain final
...
+#ifdef HAVE_PPLITE
+    // Helper for allocating/initializing PPLite's manager.
+    auto pplite_manager_alloc
+      = [](const char* domain_name) {
+          // Polyhedra are topologically closed (strict == false).
+          auto man = ap_pplite_poly_manager_alloc(false);
+          // Select the specific PPLite domain.
+          ap_pplite_poly_manager_set_kind(man, domain_name);
+          // Widening arguments satisfy the inclusion hypothesis.
+          ap_pplite_poly_manager_set_widen_spec(man, "risky");
```
caballa, line 124: I think this option is not available in https://github.com/ezaffanella/PPLite/raw/main/releases/pplite-0.7.1.tar.gz
I get the error:

```
/Users/jorge/Repos/crab-ppl/include/crab/domains/apron_domains.hpp:124:11: error: use of undeclared identifier 'ap_pplite_poly_manager_set_widen_spec'; did you mean 'ap_pplite_poly_manager_set_kind'?
          ap_pplite_poly_manager_set_widen_spec(man, "risky");
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          ap_pplite_poly_manager_set_kind
```

include/crab/domains/apron_domains.hpp
```diff
@@ -1608,6 +1642,29 @@ class apron_domain final
...
+#ifdef HAVE_PPLITE
+    // Ad-hoc handling for PSet (which is disjunctive).
+    if (ApronDom == APRON_PPLITE_PSET) {
+      if (is_bottom()) {
+        return disjunctive_linear_constraint_system_t(true /*is_false*/);
+      } else if (is_top()) {
+        return disjunctive_linear_constraint_system_t(false /*is_false*/);
+      } else {
+        auto pset = &*m_apstate;
+        auto num_disj = ap_pplite_poly_num_disjuncts(s_apman, pset);
```
caballa, line 1654: I think same problem than above. `ap_pplite_poly_num_disjuncts` does not seem to be available by the pplite 0.7.1

tests/domains/pplite_domains.cc
```diff
@@ -0,0 +1,383 @@
...
+    }
+    delete cfg;
+  }
+
+  {
+    variable_factory_t vfac;
+    z_cfg_t *cfg = prog5(vfac);
+    crab::outs() << *cfg << "\n";
+    {
+      z_poly_pplite_domain_t init;
```
caballa, line 317: You probably mean here `z_pk_apron_domain_t`

tests/domains/pplite_domains.cc
```diff
@@ -0,0 +1,383 @@
...
+      run(cfg, cfg->entry(), init, false, 1, 2, 20, stats_enabled);
+    }
+    {
+      z_poly_pplite_domain_t init;
```
caballa, line 321: Running this test (`prog5`) with `z_poly_pplite_domain_t` causes an error in crab.
The problem seems that when I called ` ap_abstract0_bound_dimension` to get an interval, the lower bound is a scalar of  type `AP_SCALAR_MPQ` and the upper bound is a scalar of type `AP_SCALAR_DOUBLE`. Is this possible? I can handle this case but current implementation assumes that both lower and upper bound have the same scalar type.

caballa (commented):
This is great. Sorry it took a while but I wanted to try by myself and run the pplite tests. I've added some comments  from the problems I found.

cmake/FindFLINT.cmake
gretadolcetti: Fixed in commit 8891857.

cmake/download_pplite.cmake
gretadolcetti: Fixed in commit 2f534b7.

include/crab/domains/apron_domains.hpp
gretadolcetti: It should be available now (update to PPLite version 0.8 in commit d3100f7).

include/crab/domains/apron_domains.hpp
gretadolcetti: It should be available now (update to PPLite version 0.8 in commit d3100f7).

tests/domains/pplite_domains.cc
ezaffanella: You are right, this was the problem (fixed in the Apron wrapper fo PPLite 0.8).
I think the Apron wrapper for PPL has the very same problem.
I do NOT know if Apron is really meant to allow for boundaries of different scalar types: I guess it should not, your assumption seems very reasonable.

tests/crab_dom.hpp
```diff
@@ -53,6 +53,9 @@ using z_dis_interval_domain_t = dis_interval_domain<z_number, varname_t>;
 using z_box_apron_domain_t = apron_domain<z_number, varname_t, APRON_INT>;
 using z_oct_apron_domain_t = apron_domain<z_number, varname_t, APRON_OCT>;
 using z_pk_apron_domain_t = apron_domain<z_number, varname_t, APRON_PK>;
+using z_poly_pplite_domain_t = apron_domain<z_number, varname_t, APRON_PPLITE_POLY>;
```
caballa, line 57: It needs to be protected by `#ifdef HAVE_PPLITE`

tests/crab_dom.hpp
```diff
@@ -126,6 +129,9 @@ using z_rgn_sign_constant_t = region_domain<z_rgn_sign_cst_params_t>;
 using q_interval_domain_t = interval_domain<q_number, varname_t>;
 using q_boxes_domain_t = boxes_domain<q_number, varname_t>;
 using q_pk_apron_domain_t = apron_domain<q_number, varname_t, APRON_PK>;
+using q_poly_pplite_domain_t = apron_domain<q_number, varname_t, APRON_PPLITE_POLY>;
```
caballa, line 135: It needs to be protected by `#ifdef HAVE_PPLITE`

tests/crab_inst.hpp
```diff
@@ -34,6 +34,9 @@ Z_RUNNER(crab::domain_impl::z_dis_interval_domain_t)
 Z_RUNNER(crab::domain_impl::z_box_apron_domain_t)
 Z_RUNNER(crab::domain_impl::z_oct_apron_domain_t)
 Z_RUNNER(crab::domain_impl::z_pk_apron_domain_t)
+Z_RUNNER(crab::domain_impl::z_poly_pplite_domain_t)
```
caballa, line 38: It needs to be protected by `#ifdef HAVE_PPLITE`

tests/crab_inst.hpp
```diff
@@ -55,6 +58,9 @@ Z_RUNNER(crab::domain_impl::z_rgn_sdbm_t)
 Q_RUNNER(crab::domain_impl::q_interval_domain_t)
 Q_RUNNER(crab::domain_impl::q_oct_apron_domain_t)
 Q_RUNNER(crab::domain_impl::q_pk_apron_domain_t)
+Q_RUNNER(crab::domain_impl::q_poly_pplite_domain_t)
```
caballa, line 64: It needs to be protected by `#ifdef HAVE_PPLITE`

caballa (commented):
all the previous are gone! just one more minor issue and ready to go!

tests/crab_inst.hpp
gretadolcetti: Fixed in e622c5d.

tests/crab_inst.hpp
gretadolcetti: Fixed in e622c5d.

tests/crab_dom.hpp
gretadolcetti: Fixed in e622c5d.

tests/crab_dom.hpp
gretadolcetti: Fixed in e622c5d.

PR #54, opened 2023-01-18
feat(inter-analysis): integreate dsa info into callsite information

LinerSu:
Giving a method to group region variables passed/returned between caller and callee based on the DSA intrinsics. 
The information is stored in the `callsite_info` class for further abstract operations of calls.
For instance, from **caller to callee**
- The caller input parameter lists: `(a, b, V1, V2, V3, V4, V5)`
- The callee input parameter lists: `(c, d, V7, V8, V9, V10, V11)`
- Intrinsics in crabIR:
```lua
crab_intrinsic(regions_from_memory_object, V7, V8, V9);
crab_intrinsic(regions_from_memory_object, V10, V11);
```
The information to group region variables is `{{V1, V2, V3}, {V4, V5}}` for the caller's inputs and `{{V7, V8, V9}, {V10, V11}}` for the callee's inputs.


include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -1066,8 +1164,19 @@ class top_down_inter_transformer final
       // If we do not analyze precisely recursive functions then we
       // must start the analysis of a recursive procedure without
       // propagating from caller to callee (i.e., top).
-      domains::callsite_info<variable_t> callsite {cs.get_func_name(), cs.get_args(), cs.get_lhs(),
-						   fdecl.get_inputs(), fdecl.get_outputs()};
+      std::vector<std::vector<variable_t>> formal_cls, actual_cls;
+      group_regions_by_dsa_intrinsics(callee_cfg, cs.get_args(),
```
caballa: This should be computed only once. We should compute it before the whole analysis starts, cache it and then just use it at each call. It should be cached in `m_ctx` similar to `get_widening_set()`.

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -1032,6 +1032,104 @@ class top_down_inter_transformer final
...
+    CRAB_LOG("inter-group", errs() << "Outs: ";
+             print_var_vector(errs(), tgt_params); errs() << "\n";);
+
+    // Step 1. get all dsa intrinsic statements
+    const basic_block_t &entry_bb = tgt_cfg.get_node(tgt_cfg.entry());
+    std::vector<intrinsic_statement_t> intrinsic_stmts;
+    intrinsic_stmts.reserve(entry_bb.size());
+    for (auto const &stmt : entry_bb) {
+      if (stmt.is_intrinsic()) {
+        auto intrinsic_stmt = dynamic_cast<const intrinsic_statement_t &>(stmt);
```
caballa: should be `static_cast` since we know it cannot fail and we don't want to rely on RTTI

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -1032,6 +1032,104 @@ class top_down_inter_transformer final
     // crab::CrabStats::count("Interprocedural.num_calling_contexts");
   }
 
+  /// @brief group regions variables in parameter lists based on target's dsa
```
caballa: What do source and target mean? Is it source = caller and target = callee ? If yes, use plese caller/callee

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -1032,6 +1032,104 @@ class top_down_inter_transformer final
...
+        }
+      }
+    }
+
+    // Step 2. construct the equivalence classes for grouping input/output
+    // regions Note that, the order of the output regions in the parameter list
+    // is the same as the output list in the dsa intrinsic
+    // The worst case of the following loop is O(m * n) where m is the number of
+    // parameters and n is the number of dsa intrinsic statements
+    unsigned int index = 0, param_lst_sz = src_params.size();
```
caballa: if this is just executed once (before the analysis starts) performance might not be an issue but the current algorithm is far from ideal. As you mention, for each argument we go over all intrinsics. This is completely unnecessary.  I would process the intrinsics only once and put them in some sort of union-find. Then, for each argument in the caller you get the argument in the callee and then you can  query the union-find to see all the elements in the equivalence class. Note that you don't a super fancy union-find implementation. You don't even need to write a separate union-find class. Just have an internal map from a variable to its parent and do standard path-compression when you do the queries.

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -419,6 +420,13 @@ class global_context {
   using func_fixpoint_map_t =
       std::unordered_map<callgraph_node_t, func_fixpoint_map_entry_t>;
 
+  using variable_vector_t = std::vector<variable_t>;
+  using region_equiv_class_t = std::vector<variable_vector_t>;
+  using callsite_t = typename cfg_t::basic_block_t::callsite_t;
+  using callsite_info_t = domains::callsite_info<variable_t>;
+  using callsite_dsa_info_map_t =
+      crab::cfg::callsite_or_fdecl_map<callsite_t, callsite_info_t>;
```
caballa: Use `unordered_map<callsite_t*, callsite_info_t>`

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -419,6 +420,13 @@ class global_context {
   using func_fixpoint_map_t =
       std::unordered_map<callgraph_node_t, func_fixpoint_map_entry_t>;
 
+  using variable_vector_t = std::vector<variable_t>;
+  using region_equiv_class_t = std::vector<variable_vector_t>;
+  using callsite_t = typename cfg_t::basic_block_t::callsite_t;
+  using callsite_info_t = domains::callsite_info<variable_t>;
+  using callsite_dsa_info_map_t =
```
caballa: Just call it `callsite_info_map_t`. I want to avoid the mention to "dsa". The idea is to keep information about the callsite. For now, we only care about one dsa-specific intrinsic but this might contain non-dsa stuff in the future

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -429,6 +437,8 @@ class global_context {
 
   // -- widening points of the callgraph (extracted from *all* entries' WTOs)
   std::set<callgraph_node_t> m_widening_set;
+  // -- map each call site statement to static callsite info
+  callsite_dsa_info_map_t m_dsa_info_map;
```
caballa: Call it `m_callsite_info_map`

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -587,10 +597,34 @@ class global_context {
 
   std::set<callgraph_node_t> &get_widening_set() { return m_widening_set; }
 
+  const callsite_dsa_info_map_t &get_dsa_info_map() const {
```
caballa: call it `get_callsite_info_map`

include/crab/analysis/inter/top_down_inter_analyzer.hpp
```diff
@@ -931,6 +965,160 @@ analyze_function(CallGraphNode cg_node,
...
+// The dsa info will be stored inside an object of class domains::callsite_info
+// for each callsite statement.
+// This analysis is performed before the inter-procedural analysis.
+// input: global_context
+// a. iterate all cfg inside wto of cfgs
+// b. for each cfg, visit callsite statement
+// c. for each callsite statement, get callee'cfg
+// d. store callsite info
+template <typename CallGraph, typename GlobalCtx>
+void analyze_callsite(const CallGraph &cg, GlobalCtx &ctx) {
```
caballa: Don't call it `analyze_callsite`. This gives the impression that the callsite is actually analyzed by the abstract interpreter. Call it something like `extract_callsite_info` or similar

include/crab/domains/inter_abstract_operations_callsite_info.hpp
```diff
@@ -17,18 +17,30 @@ template <class Variable> class callsite_info {
   const std::vector<Variable> &m_caller_out_params;
   const std::vector<Variable> &m_callee_in_params;
   const std::vector<Variable> &m_callee_out_params;
+  const std::vector<std::vector<Variable>> &m_caller_in_groups;
```
caballa: I wonder if there is something better than a vector of vectors. 
I know that you want to have explicitly each equivalence class.
But it doesn't look to me very elegant but maybe that's just personal taste. 
An alternative would be just to keep a map from Variable to Variable (similar to what an union-find would keep internally). Eg.., `{{x,y}{p,q}}` can be represented with the map `{x -> x, y -> x, p -> p, q -> p}`. This map is good to answer questions about whether two variables are in the same equivalence class but you would need to create the vector of vectors probably anyway.

LinerSu: I agree the map could help us identify whether two variables belong to the same equivalence class. However, I don't need this map for the `extend` and `restrict` methods since an object copy is performed by copying subdomains and renaming them through a vector of variables. Using a vector of vectors is the simplest way to represent equivalence classes. I prefer maintaining a vector of vectors at this point.

include/crab/analysis/inter/top_down_inter_analyzer.hpp
LinerSu: Since we passed an object of `callsite_t` by constant reference, all I can do is using `std::unordered_map<const callsite_t*, callsite_info_t>`.

caballa: Yes, that's perfect

caballa:

PR #55, opened 2023-04-18
feat(domain): add missing parameterized constructor

LinerSu:
One parameterized constructor for the product domain needs to be included, which should take a constant reference for each subdomain.

### DIFF
--- a/include/crab/domains/combined_domains.hpp
+++ b/include/crab/domains/combined_domains.hpp
@@ -70,6 +70,15 @@ public:
     }
   }
 
+  basic_domain_product2(const Domain1 &first, const Domain2 &second,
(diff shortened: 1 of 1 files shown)


caballa:
I like to have in this case only the move constructor. Any reason for this? You can make a copy and move the abstract states which is what your constructor will do

LinerSu:
I sometimes need to construct a new product domain object by taking a subdomain which is a constant reference. For example,
```c++
const product_domain_t *prod_ref = ...;
const Domain1 &dom_val1 = prod_ref->fist();
Domain2 dom_val2 = prod_ref->second();
...
product_domain_t output(dom_val1, dom_val2);
```
But I agree with your solution, copy and move is an alternative solution.

PR #59, opened 2023-07-03
PPLite native

ezaffanella:
Added native interface for PPLite library domains (mostly based on previous work by @matteboro).

Notes:
- PPLite native wrapper does not depend on Apron, hence it is compatible with ELINA wrapper;
- PPLite native wrapper in under GPL license.


caballa:
@ezaffanella : I'm going to close it because the PR is too old (too many conflicts with current `dev` branch). It's totally my fault. Instead, I created a new branch `pplite-native` with the relevant code from your PR. 

In this branch, there are two ways of using PPLite:
- cmake option `-DCRAB_USE_PPLITE=ON` that uses Apron interface
- cmake option `-DCRAB_USE_PPLITE_NATIVE=ON` that uses native interface

Both options are compatible. Eventually, we will remove the Apron interface.

PR #66, opened 2026-06-13
Add a new equality domain over symbolic variables

LinerSu:
## Summary

Adds `symbolic_variable_equality_domain`, a union-find-based abstract domain
that tracks which variables are known to hold *equal* values, without tracking
what those values actually are. Two variables are considered equal in the
concrete semantics whenever the domain has assigned them the same **symbolic
representative** (`#var0`, `#var1`, ...).

This is useful as a lightweight reduction/companion domain — e.g. for
allocation-site abstraction, where many variables end up aliasing the same
memory object and we just need to know "these are the same" without paying
for a full relational domain.

## How it works

Internally the domain is a union-find (disjoint-set) over program variables.
Each equivalence class is tagged with a **symbolic representative** — a fresh
id handed out by `term_op_val_generator_t` — wrapped in a copy-on-write holder
so copies of the state share it until one is mutated. The class stores only
this representative; there is no separate base-domain payload.

```cpp
symbolic_variable_equality_domain<Number, VariableName> dom1, dom2;

// dom1: {v1, v2} share representative #var0
dom1.set(v1, idom1);
dom1.add(v1, v2);

// dom2: {v4, v2} share representative #var1
dom2.set(v4, idom2);
dom2.add(v4, v2);
```

- **Join** (`dom1 | dom2`) keeps only the equalities both sides agree on —
  here `dom1` and `dom2` don't share any pair of equal variables, so the
  join is `{}` (top).
- **Meet** (`dom1 & dom2`) unions the equalities — `v1 ~ v2` (from `dom1`)
  and `v4 ~ v2` (from `dom2`) combine transitively into a single class
  `{v1, v2, v4} => #var0`.
- `<=` checks whether one side's equivalence classes are refinements of the
  other's (i.e. every class on the left is a subset of some class on the
  right).

A default-constructed state is **top** (no equalities known); `normalize()`
drops singleton classes (which carry no equality), and `DomainParams` toggles
that normalization, optional internal consistency checks, and inter-procedural
transformers.

## What's included

- `symbolic_variable_eq_domain.hpp`: the domain itself (`set`, `add`,
  `forget`, `project`, `normalize`, join/meet/widening/narrowing,
  `to_linear_constraint_system`, etc.)
- `term_operators.hpp`/`.cpp`: `term_op_val_generator_t`, a small counter
  that hands out fresh symbolic ids (`#varN`) for new equivalence classes.
- `unittests-symb_var_eq.cc`: a self-checking unit test that verifies results
  through the public lattice interface and validates the internal
  representation after every operation.

This is split out as its own PR because it's a general-purpose domain with no
dependency on anything object-related; it's also a dependency of an upcoming
`object_domain` PR, which will be stacked on top of this one.

## Test plan
- [x] `cmake --build . --target unittests-symb_var_eq` builds cleanly
- [x] `./test-bin/unittests-symb_var_eq` runs to completion (exit 0)

### DIFF
--- a/include/crab/domains/symbolic_variable_eq_domain.hpp
+++ b/include/crab/domains/symbolic_variable_eq_domain.hpp
@@ -122,6 +122,17 @@ void print_unordered_map(crab::crab_os &o,
 class SVEQDefaultParams {
 public:
   enum { implement_inter_transformers = 0 };
+  // when non-zero, domain operations validate their internal representation
+  // (see check_lattice_val). Off by default so normal analyses pay nothing.
+  enum { check_lattice_val = 0 };
+};
+
+// Like SVEQDefaultParams but with internal consistency checks enabled. Use for
+// unit testing, or pass it when debugging a specific analysis.
+class SVEQCheckedParams {
+public:
+  enum { implement_inter_transformers = 0 };
+  enum { check_lattice_val = 1 };
 };
 
 #define SVEQ_DOMAIN_SCOPED_STATS(NAME) CRAB_DOMAIN_SCOPED_STATS(this, NAME, 1)
@@ -342,6 +353,7 @@ private:
       }
     }
     res.normalize();
+    res.check_lattice_val();
     CRAB_LOG("symb-var-eq-join", res.dump(); crab::outs() << "\n");
     return res;
   }
@@ -424,6 +436,7 @@ private:
       }
     }
     res.normalize();
+    res.check_lattice_val();
     CRAB_LOG("symb-var-eq-meet", res.dump(); crab::outs() << "\n");
     return res;
   }
@@ -445,6 +458,63 @@ private:
 
     m_parents.insert({v, v});
     m_classes.insert({v, equivalence_class_t(val)});
+  }
+
+  // Consistency check of the internal representation. Enabled via
+  // DomainParams::check_lattice_val (e.g. SVEQCheckedParams); compiles to a
+  // no-op otherwise. Validates the invariants linking m_val, m_parents and
+  // m_classes.
+  void check_lattice_val() const {
+    if (!DomainParams::check_lattice_val) {
+      return;
+    }
+    switch (m_val) {
+    case lattice_val::bottom:
+      if (!m_parents.empty() || !m_classes.empty()) {
+        CRAB_ERROR(domain_name(),
+                   "::check_lattice_val: bottom must have empty maps");
+      }
+      break;
+    case lattice_val::top:
+      if (!m_parents.empty() || !m_classes.empty()) {
+        CRAB_ERROR(domain_name(),
+                   "::check_lattice_val: top must have empty maps");
+      }
+      break;
+    case lattice_val::neither_top_nor_bot:
+      if (m_parents.empty() && m_classes.empty()) {
+        CRAB_ERROR(domain_name(), "::check_lattice_val: neither-top-nor-bottom "
+                                  "must not be empty (it should be top)");
+      }
+      if (!m_parents.empty() && m_classes.empty()) {
+        CRAB_ERROR(domain_name(), "::check_lattice_val: m_parents is non-empty "
+                                  "but there are no equivalence classes");
+      }
+      // every element points to a representative that owns an equivalence class
+      // and every representative is self-parented
+      for (auto &kv : m_parents) {
+        const element_t &rep = kv.second;
+        auto pit = m_parents.find(rep);
+        if (pit == m_parents.end() || !(pit->second == rep)) {
+          CRAB_ERROR(domain_name(), "::check_lattice_val: representative ", rep,
+                     " of ", kv.first, " is not self-parented in m_parents");
+        }
+        if (m_classes.find(rep) == m_classes.end()) {
+          CRAB_ERROR(domain_name(), "::check_lattice_val: representative ", rep,
+                     " has no equivalence class in m_classes");
+        }
+      }
+      // every equivalence class is keyed by a self-parented representative
+      for (auto &kv : m_classes) {
+        const element_t &rep = kv.first;
+        auto pit = m_parents.find(rep);
+        if (pit == m_parents.end() || !(pit->second == rep)) {
+          CRAB_ERROR(domain_name(), "::check_lattice_val: class key ", rep,
+                     " is not a representative in m_parents");
+        }
+      }
+      break;
+    }
   }
 
   void print_elems_vector(crab::crab_os &o,
@@ -563,6 +633,7 @@ public:
       equivalence_class_t &ec_x = m_classes.at(rep_x);
       ec_x.set_absval(absval_ptr);
     }
+    check_lattice_val();
   }
 
   /// @brief get the domain value stored in current equivalent class
@@ -644,10 +715,11 @@ public:
   /// @brief Add y into the equivalence class of x
   /// @param x an element that may exist
   /// @param y an element that may exist
-  /// @note  if x does not exist, nothing changes.
+  /// @note  if x does not exist, a fresh class {x} is created first.
   ///        if y \in some cls, forget it from cls, add y into x's class
+  ///        adding an equality to top creates a class (top is not absorbing).
   void add(const element_t &x, const element_t &y) {
-    if (is_bottom() || is_top() || x == y) {
+    if (is_bottom() || x == y) {
       return;
     }
     if (!contains(x)) {
@@ -658,13 +730,15 @@ public:
     }
     element_t rep_x = find(x);
     m_parents.insert({y, rep_x});
+    check_lattice_val();
   }
   /**------------------ End union find APIs ------------------**/
 
   /**------------------ Begin domain APIs ------------------**/
-  // empty union-find
-  symbolic_variable_equality_domain(
-      lattice_val val = lattice_val::neither_top_nor_bot)
+  // A default-constructed domain is top (the empty union-find: no equalities
+  // known). It transitions to neither_top_nor_bot once an equality is recorded
+  // (via make_set, reached from set/add).
+  symbolic_variable_equality_domain(lattice_val val = lattice_val::top)
       : m_val(val) {}
   symbolic_variable_equality_domain(const this_domain_t &o) = default;
 
@@ -700,6 +774,8 @@ public:
 
   bool operator<=(const this_domain_t &e) const override {
     SVEQ_DOMAIN_SCOPED_STATS(".leq");
+    check_lattice_val();
+    e.check_lattice_val();
     if (is_bottom() || e.is_top()) {
       // _|_ <= e || this <= top
       return true;
@@ -862,7 +938,12 @@ public:
       }
       m_parents.erase(v);
     }
+    if (m_parents.empty()) {
+      // no equalities left: collapse back to top
+      set_to_top();
+    }
     normalize();
+    check_lattice_val();
     CRAB_LOG("symb-var-eq", crab::outs() << "After forget "; dump(););
   }
 
@@ -923,8 +1004,13 @@ public:
       equivalence_class_t ec = m_classes.at(rep);
       res.m_classes.insert({new_rep, ec});
     }
+    if (!res.m_parents.empty()) {
+      // res was default-constructed as top; mark it as a real state
+      res.set_neither_top_or_bottom();
+    }
     std::swap(res, *this);
     normalize();
+    check_lattice_val();
     CRAB_LOG("symb-var-eq", crab::outs() << "After projection "; dump(););
   }
 
@@ -988,6 +1074,7 @@ public:
     }
     std::swap(m_classes, new_classes);
 
+    check_lattice_val();
     CRAB_LOG("symb-var-eq", crab::outs() << "After renaming "; dump(););
   }
 
--- a/include/crab/domains/symbolic_variable_eq_domain.hpp
+++ b/include/crab/domains/symbolic_variable_eq_domain.hpp
@@ -9,6 +9,7 @@
 #include <boost/optional.hpp>
 
 #include <algorithm>
+#include <limits>
 #include <memory>
 #include <unordered_map>
 #include <vector>
@@ -18,18 +19,35 @@ namespace domains {
 
 namespace symbolic_variable_equality_domain_impl {
 // Identifier of an equivalence class. Two variables are known equal when their
-// classes carry the same id. Ids only need to be distinct within a single
-// state. We tag classes with ids (rather than a representative variable) so the
-// same value identity can be referenced across domains (e.g. object_domain's
-// field/register equalities).
+// classes carry the same id. Tagging classes with ids (rather than a
+// representative variable) is a design choice: it allows an equality relation
+// to be split across several domain values, where setting the same id in two
+// values records an equality between their elements. Because ids may travel
+// between values, they must be globally unique -- an id may appear in two
+// places only because it was deliberately copied there. A per-value allocator
+// cannot provide this (two values can allocate the same id independently),
+// hence the single shared counter below.
 using class_id_t = uint32_t;
 
-// Allocate a globally-fresh class id from a self-contained monotonic counter
-// shared across all domain instances. Assumes single-threaded use.
-inline class_id_t fresh_class_id() {
+// The single, monotonically-increasing source of class ids, shared by all
+// domain instances. Assumes single-threaded use.
+inline class_id_t &class_id_counter() {
   static class_id_t counter = 0;
+  return counter;
+}
+
+// Produce a fresh class id, strictly greater than every id produced before
+// and therefore different from every id in use anywhere.
+inline class_id_t fresh_class_id() {
+  class_id_t &counter = class_id_counter();
+  if (counter == std::numeric_limits<class_id_t>::max()) {
+    CRAB_ERROR("symbolic_variable_equality_domain: class id counter overflow");
+  }
   return counter++;
 }
+
+// True iff `id` was produced by fresh_class_id() at some point.
+inline bool is_valid_class_id(class_id_t id) { return id < class_id_counter(); }
 } // namespace symbolic_variable_equality_domain_impl
 
 class SVEQDefaultParams {
@@ -160,22 +178,6 @@ private:
       }
     }
     return boost::none;
-  }
-
-  static class_id_t fresh_class_id() {
-    return symbolic_variable_equality_domain_impl::fresh_class_id();
-  }
-
-  // Allocate a fresh id guaranteed distinct from every class currently in this
-  // state, so each equivalence class keeps a unique id. The counter is
-  // monotonic, so this only loops if an explicitly-assigned id happens to fall
-  // in the counter's range.
-  class_id_t get_fresh_unused_class_id() const {
-    class_id_t id = fresh_class_id();
-    while (rep_with_class_id(id)) {
-      id = fresh_class_id();
-    }
-    return id;
   }
 
   /// @brief Build a map from representative to an ordered set with all the
@@ -358,13 +360,19 @@ private:
                      " has no equivalence class in m_classes");
         }
       }
-      // every equivalence class is keyed by a self-parented representative
+      // every equivalence class is keyed by a self-parented representative and
+      // carries an id that the factory actually produced
       for (auto &kv : m_classes) {
         const element_t &rep = kv.first;
         auto pit = m_parents.find(rep);
         if (pit == m_parents.end() || !(pit->second == rep)) {
           CRAB_ERROR(domain_name(), "::check_lattice_val: class key ", rep,
                      " is not a representative in m_parents");
+        }
+        if (!symbolic_variable_equality_domain_impl::is_valid_class_id(
+                kv.second)) {
+          CRAB_ERROR(domain_name(), "::check_lattice_val: class id #id",
+                     kv.second, " was never produced by fresh_class_id()");
         }
       }
       break;
@@ -482,9 +490,16 @@ public:
     return it->second;
   }
 
+  /// @brief Produce a fresh class id, different from every id ever produced.
+  /// @note  Ids passed to set() must originate from here (either directly or
+  ///        read back from a class via get_class_id()).
+  static class_id_t fresh_class_id() {
+    return symbolic_variable_equality_domain_impl::fresh_class_id();
+  }
+
   /// @brief tag x's class with class id \p id
   /// @param x an element
-  /// @param id a class id
+  /// @param id a class id previously produced by fresh_class_id()
   /// @note  if x is new and no class already holds \p id, create a class {x}
   ///        with \p id;
   ///        if some class already holds \p id, x's class is merged into it --
@@ -493,6 +508,12 @@ public:
   void set(const element_t &x, class_id_t id) {
     if (is_bottom()) {
       return;
+    }
+    if (!symbolic_variable_equality_domain_impl::is_valid_class_id(id)) {
+      // an id the factory never produced could later be handed out by
+      // fresh_class_id(), silently merging unrelated classes
+      CRAB_ERROR(domain_name(), "::set: class id #id", id,
+                 " was never produced by fresh_class_id()");
     }
 
     if (!contains(x)) {
@@ -586,11 +607,10 @@ public:
       return;
     }
     if (!contains(x)) {
-      // Create an isolated fresh class for x with a symbol distinct from every
-      // existing class. Going through set() with a plain fresh id could instead
-      // fold x into an unrelated class if that id collides with an explicitly-
-      // assigned symbol already in the state.
-      try_make_set_by_raw_val(x, get_fresh_unused_class_id());
+      // Create an isolated fresh class for x. The factory is monotonic and
+      // set() only accepts ids the factory produced, so a fresh id is greater
+      // than every id in use and cannot land on an existing class.
+      try_make_set_by_raw_val(x, fresh_class_id());
     }
     element_t rep_x = find(x);
     if (!contains(y)) {
--- a/include/crab/domains/symbolic_variable_eq_domain.hpp
+++ b/include/crab/domains/symbolic_variable_eq_domain.hpp
@@ -4,7 +4,6 @@
 #include <crab/domains/abstract_domain_specialized_traits.hpp>
 #include <crab/domains/backward_assign_operations.hpp>
 #include <crab/domains/constant.hpp>
-#include <crab/domains/term/term_operators.hpp>
 #include <crab/support/stats.hpp>
 
 #include <boost/optional.hpp>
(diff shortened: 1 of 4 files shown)


include/crab/domains/term/term_operators.hpp
```diff
@@ -54,6 +54,14 @@ term_operator_t conv2termop(bool_operation_t op);
 boost::optional<arith_operation_t> conv2arith(term_operator_t op);
 boost::optional<bitwise_operation_t> conv2bitwise(term_operator_t op);
 boost::optional<bool_operation_t> conv2bool(term_operator_t op);
+
+class term_op_val_generator_t {
+private:
+  static uint32_t m_next_val;
+
+public:
+  static uint32_t get_next_val() { return m_next_val++; }
```
caballa: I would really like to avoid using a global counter.

caballa: Also not sure this is the place

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1110 @@
+#pragma once
+
+#include <crab/domains/abstract_domain.hpp>
+#include <crab/domains/abstract_domain_specialized_traits.hpp>
+#include <crab/domains/backward_assign_operations.hpp>
+#include <crab/domains/constant.hpp>
+#include <crab/domains/symbolic_variable_eq_domain.hpp>
```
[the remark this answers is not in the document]
LinerSu: Fixed in 4107a7f7 — removed the redundant self-include.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1110 @@
...
+    }
+    if (!contains(x)) {
+      set(x, this_domain_t::__get_fresh_symb_var());
+    }
+    if (contains(y)) {
+      *this -= y;
+    }
+    element_t rep_x = find(x);
+    m_parents.insert({y, rep_x});
+  }
```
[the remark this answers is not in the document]
LinerSu: Fixed in 33c74072 — a default-constructed domain is now top, and add() no longer early-returns on top, so adding an equality to top creates a class (top is not absorbing). Covered by a new unit-test case.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1110 @@
...
+              {new_v, (*parent_old_v == old_v ? new_v : *parent_old_v)});
+          auto it = m_classes.find(old_v);
+          if (it != m_classes.end()) {
+            // update m_classes
+            equivalence_class_t ec = it->second;
+            m_classes.erase(it);
+            m_classes.insert({new_v, std::move(ec)});
+          }
+        }
+      }
```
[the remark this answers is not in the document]
LinerSu: Fixed in 577c66ab — rewrote rename() to first build the old->new substitution, then rewrite m_parents and m_classes in a single pass each, so nothing is inserted into a map while it is being iterated.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1110 @@
...
+
+  void caller_continuation(const callsite_info<variable_t> &callsite,
+                           const this_domain_t &callee) override {
+    SVEQ_DOMAIN_SCOPED_STATS(".caller_cont");
+    inter_abstract_operations<
+        this_domain_t,
+        DomainParams::implement_inter_transformers>::callee_entry(callsite,
+                                                                  callee,
+                                                                  *this);
+  }
```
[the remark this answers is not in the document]
LinerSu: Fixed in 30bc907c — caller_continuation now delegates to inter_abstract_operations::caller_continuation.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1110 @@
...
+/// @tparam BaseDomain the type of the domain value that represents the base
+/// domain in the object domain
+/// @tparam DomainParams domain parameter for inter-procedural analysis
+template <class BaseDomain, class DomainParams = SVEQDefaultParams>
+class symbolic_variable_equiality_domain final
+    : public abstract_domain_api<
+          symbolic_variable_equiality_domain<BaseDomain, DomainParams>> {
+public:
+  using symb_eq_domain_t =
+      symbolic_variable_equiality_domain<BaseDomain, DomainParams>;
```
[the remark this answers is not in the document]
LinerSu: Fixed in 4107a7f7 — renamed to symbolic_variable_equality_domain / _impl (equality) throughout.

tests/domains/unittests-symb_var_eq.cc
```diff
@@ -0,0 +1,346 @@
...
+    dom9.add(v3, v4);
+
+    dom10.set(v1, idom3);
+    dom10.add(v1, v2);
+    /*
+    join: {[v1,v2]=>#var0}
+    dom9 <= dom10
+    meet: {[v1,v2]=>#var0,[v3,v4]=>#var1}
+    */
+    perfrom_domain_operations(dom1, dom2);
```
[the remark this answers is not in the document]
LinerSu: Fixed in 6d2a7271, and the test was subsequently rewritten in 21c17a0e to be self-checking (it asserts expected results rather than only printing), which removes this whole class of mistake.

tests/domains/unittests-symb_var_eq.cc
```diff
@@ -0,0 +1,346 @@
...
+
+    /*
+    join: {[v2,v3]=>#var0}
+    no relation between dom1 and dom2
+    meet: {[v1,v2,v3,v4,v10,v11,v12]=>#var0}
+    */
+    perfrom_domain_operations(dom1, dom2);
+    test_domain_t dom5(dom1);
+    dom5 -= v10;
+    crab::outs() << "After forgetting " << v2 << " in Dom 1: " << dom5 << "\n";
```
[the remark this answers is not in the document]
LinerSu: Fixed in 6d2a7271; the test was then rewritten in 21c17a0e to be self-checking, so output messages are no longer hand-written and cannot drift from the code.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1110 @@
...
+    SVEQ_DOMAIN_SCOPED_STATS(".narrowing");
+    return this->operator&(e);
+  }
+
+  this_domain_t
+  widening_thresholds(const this_domain_t &abs,
+                      const thresholds<number_t> &ts) const override {
+    SVEQ_DOMAIN_SCOPED_STATS(".widening");
+    return *this;
+  }
```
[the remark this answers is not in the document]
LinerSu: Fixed in 30bc907c — widening_thresholds now returns operator|(abs) (join). This domain has finite height so widening = join, and the result is a proper upper bound of both operands.

LinerSu:
## Update: review feedback addressed + hardening beyond the original PR

All 8 Copilot review comments are fixed and resolved. Beyond those, this branch also strengthens the domain and its test. Summary of the new commits on top of the original:

### Review fixes
- **typo** `equiality` → `equality` (public type names) and removed a redundant self-include. (`4107a7f7`)
- **`rename()` UB**: rewrote to build an old→new substitution and rewrite `m_parents`/`m_classes` in a single pass each — no insertion into a map mid-iteration, and O(n+m) instead of O(k·n). (`577c66ab`)
- **`widening_thresholds`** now returns `operator|(abs)` (join) instead of `*this`; **`caller_continuation`** delegates to `caller_continuation` (was `callee_entry`). (`30bc907c`)
- **`add` on top**: see "top semantics" below. (`33c74072`)

### Correctness / semantics hardening
- **Top is now the default state** and is no longer absorbing for `add`: `top.add(x,y)` creates `{x,y}`. `forget` that empties the maps collapses back to top, and `project` marks a non-empty result as a real state, keeping the lattice flag consistent with the contents.
- **`leq`** treats a not-yet-normalized singleton-only state the same as top (`top <= e` iff `e` has no equalities), via a new `has_no_equalities()` helper.
- **`normalize()` implemented** (was a no-op): drops singleton classes, collapsing a singleton-only state to top. Gated on `DomainParams::normalize` (on by default); `SVEQNoNormalizeParams` disables it where singleton classes must be kept (e.g. field/register equalities). (`e082025b`)
- **`to_linear_constraint_system` / `to_disjunctive_linear_constraint_system` implemented** (were `CRAB_ERROR` stubs): emit `member == representative` for each non-singleton class. (`f7ac13ac`)

### API / cleanup
- Parameterized the domain on `<Number, VariableName, DomainParams>` instead of a phantom `BaseDomain` (only its `number_t`/`varname_t` were used), mirroring `sign_domain`.
- Encapsulated `symbolic_var` (private field, `explicit` ctor, `value()` accessor, dropped the implicit `uint32_t` conversion); removed dead code (`symb_var_hash`, `set_is_absolute_complement`, an unused `operator==`, a redundant `equiv_classes_elems` overload) and an unused include; added the direct std includes.

### Test hardening
- Rewrote `unittests-symb_var_eq` to be **self-checking**: instead of only printing, it builds ground-truth reference domains and verifies results through the public lattice interface only (`dom <= make_dom({{a,b}})` to test entailment), with explicit `[PASS]/[FAIL]` reporting and a non-zero exit on failure. It also runs with an internal consistency checker (`check_lattice_val`, via `SVEQCheckedParams`) that validates the representation after every operation. 122 checks, including dedicated cases for `add`-on-top, `normalize`, and `to_linear_constraint_system`.

A new `check_lattice_val` policy and `SVEQNoNormalizeParams`/`SVEQCheckedParams` parameter classes were added so these checks/behaviors can be toggled per instantiation at zero cost when off.

tests/domains/unittests-symb_var_eq.cc
```diff
@@ -0,0 +1,483 @@
+#include "../common.hpp"
+#include "../program_options.hpp"
+
+#include <crab/domains/symbolic_variable_eq_domain.hpp>
+#include <crab/domains/intervals.hpp>
+
+#include <vector>
```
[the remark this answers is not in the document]
LinerSu: Fixed in 2dce7149 — removed the now-unused intervals.hpp include (the domain is no longer instantiated over an interval base).

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1262 @@
...
+    SVEQ_DOMAIN_SCOPED_STATS(".rename");
+
+    if (is_top() || is_bottom()) {
+      return;
+    }
+    if (old_elements.size() != new_elements.size()) {
+      CRAB_ERROR(domain_name(),
+                 "::rename with input vectors of different sizes");
+    }
+    CRAB_LOG("symb-var-eq", crab::outs() << "Renamimg ";
```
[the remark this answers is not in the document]
LinerSu: Fixed in 2dce7149 — corrected the debug log to "Renaming".

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1262 @@
...
+
+  // typedefs for equality domain
+  using element_t = variable_t;
+  using element_set_t = std::vector<element_t>;
+  using domain_t = class symbolic_variable_equality_domain_impl::symbolic_var;
+  using var_id_t = typename domain_t::var_id_t;
+  using parents_map_t = std::unordered_map<element_t, element_t>;
+  using equivalence_class_t =
+      symbolic_variable_equality_domain_impl::equivalence_class<domain_t>;
+  using equivalence_class_elems_t =
```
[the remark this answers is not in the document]
LinerSu: Updated the PR description. The BaseDomain payload was removed when the domain was reparameterized to <Number, VariableName, DomainParams> (a class now stores only its symbolic representative, copy-on-write); the description no longer claims a BaseDomain value.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1262 @@
...
+        return kv.first;
+      }
+    }
+    return boost::none;
+  }
+
+  static domain_t __get_fresh_symb_var() {
+    return symbolic_variable_equality_domain_impl::make_fresh_var_symbol<
+        symbolic_variable_equality_domain_impl::symbolic_var>();
+  }
```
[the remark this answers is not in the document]
LinerSu: Good catch — this is reachable, and we reproduced it: tag a class with an explicit symbol equal to the next counter value, then add() a fresh variable, and set()'s merge-on-equal-symbol silently folds it into that class.

Fixed in af1d868d: add() now allocates via get_fresh_unused_symb_var(), which skips any id already live in the state, so each equivalence class keeps a distinct symbol (the global counter is monotonic, so this only loops when an explicitly-assigned symbol falls in the counter's range). make_set() also asserts (under the consistency-check policy) that a new class's symbol is not already in use, and a unit-test case forces the collision and verifies add() leaves the unrelated class untouched.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+}; // end class equivalence_class
+
+/// @brief helper function for logging and debugging map
+/// @tparam Key the type of value for keys
+/// @tparam Value the type of value for values
+/// @param o crab os stream
+/// @param m unordered map object
+template <typename Key, typename Value>
+void print_unordered_map(crab::crab_os &o,
```
caballa: I think it's used just once. I would inline this definition where it's used.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+  void try_make_set_by_raw_val(const element_t &v, domain_t absval) {
+    if (!contains(v)) {
+      std::shared_ptr<domain_t> absval_ptr =
+          std::make_shared<domain_t>(std::move(absval));
+      make_set(v, absval_ptr);
+    }
+  }
+
+  boost::optional<const element_t>
+  check_symb_val_exists(const std::shared_ptr<domain_t> absval_ptr) const {
```
caballa: I would change the name. This function seems to find the representative whose equivalence class is equal to `absval_ptr`. This function is expensive and unusual so I need to understand where it is used

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+    equivalence_class_t absval_cls = equivalence_class_t(absval_ptr);
+    for (const auto &kv : m_classes) {
+      if (kv.second.value_equals(absval_cls)) {
+        return kv.first;
+      }
+    }
+    return boost::none;
+  }
+
+  static domain_t __get_fresh_symb_var() {
```
caballa: I would inline the definition of `make_fresh_var_symbol` here since it's the only used.

caballa: Why are you using `__` as prefix to the function name?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+    return symbolic_variable_equality_domain_impl::make_fresh_var_symbol<
+        symbolic_variable_equality_domain_impl::symbolic_var>();
+  }
+
+  // True iff some class in this state already stores symbol `sym`.
+  bool symb_val_in_use(const domain_t &sym) const {
```
caballa: This function is pretty similar to `check_symb_val_exists`. 
I would try to refactor code between the two and having better name as I suggested before.
I'm still wondering why we need these functions

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+      }
+    }
+    return false;
+  }
+
+  // Allocate a fresh symbol guaranteed distinct from every class currently in
+  // this state, so each equivalence class keeps a unique symbol id. The global
+  // counter is monotonic, so this only loops if an explicitly-assigned symbol
+  // happens to fall in the counter's range.
+  domain_t get_fresh_unused_symb_var() const {
```
caballa: I was very confused because you use `domain_t` to actually refer to a `symbolic_var`. Please rename `domain_t` to `symb_var_t` or something similar!

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+      s.insert(it, kv.first);
+    }
+    return res;
+  }
+
+  element_set_t get_all_members() const {
+    element_set_t out;
+    out.reserve(m_parents.size());
+    for (auto &kv : m_parents) {
+      out.push_back(kv.first);
```
caballa: You use a std::vector as `element_set_t`. In other functions, you insert in order here you don't. Just make sure that `out` is sorted

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+    element_set_t out;
+    out.reserve(m_parents.size());
+    for (auto &kv : m_parents) {
+      element_t rep = kv.second;
+      if (rep == e_rep) {
+        auto it = std::upper_bound(out.begin(), out.end(), kv.first);
+        out.insert(it, kv.first);
+      }
+    }
+    return out; // no need std::move since Return Value Optimization (RVO) is
```
caballa: Remove the comment about RVO

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+                // enabled
+  }
+
+  /// @brief a helper method for domain operation join
+  /// @param left one abstract state
+  /// @param right one abstract state
+  /// @return a new abstract state holding the joined result
+  /// @details Computes the intersection of two equivalence classes
+  /// to find the least common elements. It only computes a set of
```
caballa: These comments are not very useful. The spec is clear to me but not sure if the implementation follows that spec. The spec should be: "for any pair of variables (x,y) in the left operand such that x==y (i.e., x and y have the same rep) if (x,y) is also an equality in the right operand then add the pair to the result.". Is this what the code does?

caballa: What is `^` ? XOR?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+  /// The solution merges two equivalence classes if they share common
+  /// elements. As a result, it computes a set of equivalence classes where
+  /// each class is a superset of two or more classes from the given two
+  /// abstract states. That is, the result of the meet will keep equalities
+  /// that appear either in the left or in the right state.
+  /// By definition, our meet is an over-approximation meet of concrete.
+  /// E.g. there is no bottom state after meet.
+  /// Formally, forall cls_x \in left. forall cls_y \in right ::
+  ///           cls_x ^ cls_y != \empty => cls_x union cls_y
+  this_domain_t meet(const this_domain_t &left,
```
caballa, line 277: Why is the meet so complicated?
I would expect something like this

```
this_domain_t meet(const this_domain_t &left, const this_domain_t &right) const {
    this_domain_t res = left;
    for (auto &kv : right.m_parents) {
        const element_t &k = kv.first;
        const element_t &v = kv.second;
        if (k != v) {  
            // perhaps call first try_make_set or something like that
            res.add(k, v);
        }
    }
    res.normalize();
    res.check_lattice_val();
    return res;
}
```

caballa: Calling `get_all_members_from_an_equiv_class` seems too inefficient.

caballa, line 277: Shouldn't do first the cases where the operands are top or bottom?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+  /// equivalence classes where each class is a subset of two classes from
+  /// the given two abstract states. That is, the result of the join only
+  /// retains equalities that appear in both the left and right states.
+  /// Formally, forall cls_x \in left. forall cls_y \in right ::
+  ///           cls_x ^ cls_y
+  ///
+  /// This operation runs in quadratic time -- O(n^2).
+  /// The computation follows the structure of unordered_map.
+  /// The operation requires constructing a new unordered_map.
+  this_domain_t join(const this_domain_t &left,
```
caballa, line 223: Shouldn't do first the cases where the operands are top or bottom?

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+    }
+    for (auto &kv : m_parents) {
+      if (!(kv.first == kv.second)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  void print_elems_vector(crab::crab_os &o,
```
caballa: The pretty-printing functions look pretty similar to each other. 
I would ask claude to make it less verbose. For instance
```
template <typename Range, typename PrintFn>
void print_separated(crab_os &o, const Range &r, PrintFn fn, const char open, const char close, const char *sep = ",") {
    o << open;
    bool first = true;
    for (auto &elem : r) {
        if (!first) o << sep;
        fn(elem);
        first = false;
    }
    o << close;
}

void print_elems_vector(crab_os &o, const std::vector<element_t> &elems) const {
    print_separated(o, elems, [&](const element_t &e) { o << e; }, '[', ']');
}

void print_equiv_classes(crab_os &o, const equivalence_class_elems_t &equiv_classes,
                         bool verbose = false) const {
    print_separated(o, equiv_classes, [&](const auto &kv) {
        print_elems_vector(o, kv.second);
        if (!verbose) o << "=>" << *(m_classes.at(kv.first).get_absval());
    }, '{', '}');
}

void print_classes_vals(crab_os &o) const {
    print_separated(o, m_classes, [&](const auto &kv) {
        o << kv.first << "=>" << *(kv.second.get_absval());
    }, '{', '}');
}
```

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+  }
+};
+
+template <class SYMBVAR> SYMBVAR make_fresh_var_symbol() {
+  return SYMBVAR(term::term_op_val_generator_t::get_next_val());
+}
+
+/// @brief a copy-on-write wrapper for the domain value
+/// @tparam Domain the type of equality domain
+template <class Domain> class equivalence_class {
```
caballa: AFIK, `Domain` will be only a wrapper to an uint32. Therefore, it's unclear to me if COW optimization will pay off

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+    const element_t &rep_y = find(y);
+    return rep_x == rep_y;
+  }
+
+  /// @brief find the representative without path-compression
+  /// @param v an element in some set
+  /// @attention v must be contained in the current domain; otherwise, an
+  /// error is reported
+  /// @return returns the representative of the set that contains the element v
+  element_t find(const element_t &v) const {
```
caballa, line 475: `find` should call `find_opt` and return ERROR if the result is none. Don't duplicate the code please

include/crab/domains/symbolic_variable_eq_domain.hpp
caballa: I see now it's used in `set` as a kind of on-the-fly normalization step

caballa (commented):
My main problem is having a global counter to generate unique Ids for variables. 
```
template <class SYMBVAR> SYMBVAR make_fresh_var_symbol() {
  return SYMBVAR(term::term_op_val_generator_t::get_next_val());
}
```

and the use of `get_fresh_unused_symb_var` which is a "retry loop" because  join/meet create new states with renamed ids, and the counter can fall behind. This is a very fragile design. I would try to avoid this global counter. Why do we need it? Perhaps, are you trying to be too general? `element_t` is already `variable_t` which have already unique IDs. Can we use these IDs instead?

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Removed.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Done — find calls find_opt now.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Inlined it; the dump table now uses the shared print_separated helper too.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Dropped it — only get_all_variables used it, and nothing needs the result sorted.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Rewrote the comment with your spec, and confirmed the code does that: it walks left's equalities and keeps the ones that also hold in right.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: These helpers only run from operator| / operator& after top/bottom are handled, so they assume neither. Added a @pre note.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Factored them with a print_separated helper (your version) — the three printers and the dump table all use it now.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Simplified it: meet is now a copy of left, then union in each of right's equalities. To make that correct I also redefined add() as a real union — it used to move an element and drop its old equalities. object_domain doesn't call add, so no fallout.

LinerSu: Same here — meet runs from operator& after top/bottom are handled. Added a @pre note.

include/crab/domains/symbolic_variable_eq_domain.hpp
```diff
@@ -0,0 +1,1294 @@
...
+        add(*y_opt, x);
+      } else {
+        make_set(x, absval_ptr);
+      }
+    } else {
+      // Modify the abstract state of the whole equivalence class
+      element_t rep_x = find(x);
+      equivalence_class_t &ec_x = m_classes.at(rep_x);
+      ec_x.set_absval(absval_ptr);
+    }
```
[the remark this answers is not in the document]
LinerSu: Fixed — set() on an existing var now merges into the class that already owns that id instead of overwriting (which could leave two classes sharing an id). Uses the union add().

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Merged it with symb_val_in_use into one rep_with_class_id(id) that returns the class's representative (or none), renamed as you suggested.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Folded into rep_with_class_id — one scan, one function. It backs the 'same id => same class' merge that set()/add() use.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Dropped it — a class value is just a uint32 now (class_id_t), stored by value. No shared_ptr / equivalence_class.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Renamed to class_id_t (a uint32) — it's an equivalence-class id, not a variable.

include/crab/domains/term/term_operators.hpp
LinerSu: Moved the counter into the domain's own header (fresh_class_id, a function-local static) so the header is self-contained — no term_operators dependency. Still a monotonic counter for unique class ids; can make it injectable later if you'd prefer.

include/crab/domains/symbolic_variable_eq_domain.hpp
LinerSu: Removed — make_fresh_var_symbol is gone; classes get ids from fresh_class_id().

LinerSu:
@caballa thanks for the detailed review — addressed all of it in 52392760 and resolved the threads. The main changes:

- A class value is now a plain `uint32` (`class_id_t`) stored by value — dropped `symbolic_var`, the copy-on-write `equivalence_class`, and all `shared_ptr`.
- The fresh-id counter lives in this header now (`fresh_class_id`), so the domain no longer depends on `term_operators` (reverted that addition). It's still a monotonic counter for unique class ids — happy to make it injectable later if you'd prefer.
- Simplified `meet` (copy left, then union in each of right's equalities) and made `add` a proper union; fixed `set` to merge instead of duplicating when an id is already in use.
- Merged the two id scans into one `rep_with_class_id`, `find` now calls `find_opt`, factored the printers with a `print_separated` helper, plus the smaller comment/dead-code items.

Net ~-150 lines. Could you take another look when you get a chance?

LinerSu:
One more note on the id counter, since "avoid the global counter" is the part I'd most like feedback on:

The class id isn't just an internal tag — it's a neutral token meant to be shared *across* domains. object_domain keeps separate equality domains for fields and registers, and links them by giving related entities the same id (e.g. a field and the register that holds it). A union-find representative can't play that role: it's a variable local to one subdomain (a field variable means nothing in the register domain), so we need a separate id minted from a counter. Within one state the id only has to be unique; across the analysis the ids stay globally unique so the same value identity is comparable everywhere — which is why it's a single monotonic counter.

It's now self-contained in the header (no `term_operators`), but it is still global mutable state.

caballa:
In that case, I admit that you cannot avoid having a global id factory. The right solution would be to pass the id factory to constructors and bottom/top methods but that would be require a whole re-design of Crab. 

However, you can still avoid the use of `get_fresh_unused_class_id`. 
Take a look at `str_var_alloc_col` defined in `varname_factory.hpp`. It resolves exactly that problem.

LinerSu:
I looked at `str_var_alloc_col`. It works when the allocated names stay inside one domain value — term_domain's case. Our case is different: in object_domain the equality domain is split into several values (`m_eq_regs_dom` for registers, one per object for fields), and the **same class id is shared across values** — `ref_load`/`ref_store` set one id on both sides, and the reduction matches ids across values to recover `reg ≈ fld`.

**Our observation: with a per-value allocator, class ids will collide across values**, because each allocator only knows its own value. Example:

```
eq_regs:  {r0}@_x1, {r1}@_x0
eq_fld1:  {f0,f1}@_x1            // object 1
eq_fld2:  {f2}@_x0               // object 2

ref_load r2 := f3    (f3 in object 2, no id yet → create one)
  eq_fld2's allocator creates _x1
  set(r2, _x1) in eq_regs → finds {r0}@_x1
  → r2 merged with r0: false equality r2 ≈ r0
  → reduction also gets r0 ≈ f3, and r0,r2 ≈ f0,f1 once object 1 is cached
```

We only wanted r2 ≈ f3. A global counter gives exactly that: an id matches only where we deliberately put it. And the worst case with a global counter is a lost match (imprecision); with per-value allocators it's a wrong equality that looks like a legitimate state.

So we stay with the global counter; `get_fresh_unused_class_id` is removed in 6af6ebac (`add()` takes a fresh id from the counter directly, with two O(1) guards).

caballa (approved):

caballa:
Already merged in https://github.com/seahorn/crab/commit/eaef87979fa20dbac8e9d8e645c3c79774a68b70
I couldn't merge directly this PR because I needed to:

1) rebase `dev`
2) extend the PR so that `expected_results.out` is updated

I also squashed all the commits into one. I made sure that author was @LinerSu.

PR #67, opened 2026-07-03
Add tvpi_dbm: a template DBM (tDBM) abstract domain

LinerSu:
This PR adds `tvpi_dbm_domain`, a weakly relational abstract domain implementing
"Template DBM: A New Weakly Relational Domain for Efficient Memory-Access Validation"
(Su, Navas, Gurfinkel — VSTTE 2025).

## The domain

Given a coefficient template `T = {1, a, b, ...}` (positive integers, `1 ∈ T`), tDBM represents

    I_T = { a*x - b*y <= c  |  x, y ∈ Vars,  a, b ∈ T,  c ∈ ℤ }

    Zones (T = {1})  ⊂  tDBM (I_T)  ⊂  TVPI  ⊂  Polyhedra

Every constraint is encoded as a unit-coefficient difference constraint over *ghost
variables* `ghost(v,k) = k*v`, so any DBM-like base domain can represent it:
`tvpi_dbm_domain<BaseDom>` is parameterized over the base (instantiated with
`split_dbm_domain` in the tests).

- **Ghost variables** come from the variable factory's existing cached string API via a
  stateless manager (`tvpi/ghost_variable_manager.hpp`): `name(ghost(v,k)) = v.str() + "*" + k`;
  identity and lifetime are owned by the factory.
- **Canonical storage**: constraints are divided through by the gcd of their coefficients,
  flooring the bound (`Σ ai·xi ≤ c  ⟺  Σ (ai/g)·xi ≤ ⌊c/g⌋` over ℤ), so all constraints on a
  variable pair meet on the same ghost variables.
- **Saturation**: `normalize()` runs DbmTvpiSaturation — the Resultant rule on
  coefficient-misaligned constraint pairs (`tvpi_reduce`), with aligned transitivity performed
  by the base domain.  A per-constraint incremental variant (TvpiIncrementalSaturation) runs
  inside `assume`/`entails`/`apply`.
- The coefficient template lives in `fixed_tvpi_domain_params`, kept **sorted ascending**
  (ghost re-assignment for self-referential assignments relies on it).

## Changes outside the new files

| file | lines | what |
|---|---|---|
| `split_dbm.hpp` | +27 | `difference_bound(x,y)`: the tightest known `c` with `y − x ≤ c` — the one operation tDBM needs beyond `abstract_domain_api` (a `static_assert` in the domain states the requirement) |
| `linear_constraints.hpp` | +6 | unary `operator-` on variables |
| `abstract_domain_params.{hpp,cpp}` | +19 | sorted-ascending invariant for `fixed_tvpi_domain_params::m_coefficients` |

## Tests

`unittests-tvpi{,2,3}` cover assumptions, entailment, join and widening;
`unittests-tvpi-normalize` covers the eager `TVPIDBMNormalizeParams` variant;
`tvpi`/`tvpi2` run the domain through the intra-procedural analyzer with assertion
checking; `unittests-tvpi-reduce` is a functional audit of both saturation procedures,
one concrete Resultant derivation per case:

| derivation | incremental | `normalize()` |
|---|---|---|
| `3y−z≤0` + `2x−2y≤0` ⟹ `3x−z≤0` | ✓ | ✓ |
| `w−3x≤0` + `2x−2y≤0` ⟹ `w−3y≤0` | ✓ | ✓ |
| `w−2x≤0, 3y−z≤0` + `2x−2y≤0` ⟹ `3w−2z≤0` | ✓ | ✓ |
| `w−3x≤0, 2y−z≤0` + `2x−2y≤0` ⟹ `2w−3z≤0` | ✓ | ✓ |
| `y≤5` + `2x−3y≤0` ⟹ `x≤7` | ✓ | ✓ |
| `2y−x≤0` + `x≤7` ⟹ `y≤3` | — (see below) | ✓ |
| `3w−2z≤0` case with constraints permuted | ✓ | ✓ |

Single-variable bounds are deliberately not reduced incrementally (bounds are the most
frequent constraint; the alignment sweep would cost `O(|T|²·N)` per bound); `normalize()`
derives their consequences, and the audit test pins this as expected behavior.


LinerSu:
@caballa could you review this when you have a chance? It adds the tDBM domain from our VSTTE'25 paper on top of dev — core changes outside the new files are intentionally minimal (a difference_bound query on split_dbm, unary minus on variables, and a sortedness invariant on the fixed-TVPI coefficient template).

include/crab/domains/tvpi/ghost_variable_manager.hpp
```diff
@@ -0,0 +1,56 @@
...
+public:
+  using variable_t = Variable;
+  using varname_t = typename variable_t::varname_t;
+  using number_t = ikos::z_number;
+
+  /// Marker separating the original name from the coefficient.
+  static constexpr char marker = '*';
+
+  /// Suffix encoding of coefficient @p k, e.g. "*3".
+  static std::string encode(const number_t &k) {
```
caballa, line 45: make it private?

include/crab/domains/split_dbm.hpp
```diff
@@ -2872,6 +2872,33 @@ class split_dbm_domain final
...
+      return boost::optional<number_t>();
+    }
+    SubGraph<graph_t> g_excl(g, 0);
+    if (g_excl.elem(*ai, *bi) && g_excl.elem(*bi, *ai) &&
+        g_excl.edge_val(*ai, *bi) == Wt(0) &&
+        g_excl.edge_val(*bi, *ai) == Wt(0)) {
+      return number_t(0);
+    } else if (g_excl.elem(*ai, *bi)) {
+      return number_t(g_excl.edge_val(*ai, *bi));
+    } else if (g_excl.elem(*bi, *ai)) {
```
caballa: merge the last two cases to save the call to `elem`

include/crab/domains/tvpi/ghost_variable_manager.hpp
```diff
@@ -0,0 +1,56 @@
...
+namespace tvpi_utils {
+
+template <typename Variable> class ghost_variable_manager {
+public:
+  using variable_t = Variable;
+  using varname_t = typename variable_t::varname_t;
+  using number_t = ikos::z_number;
+
+  /// Marker separating the original name from the coefficient.
+  static constexpr char marker = '*';
```
caballa, line 42: make it private?

include/crab/domains/tvpi_dbm.hpp
```diff
@@ -0,0 +1,1967 @@
...
+  /// Guards against re-entrant calls to incremental_tvpi_reduce from within
+  /// tvpi_reduce() or another incremental pass.
+  bool m_reduce_active = false;
+
+  // ============================================================
+  // Numeric utilities
+  // ============================================================
+
+  /// Euclidean gcd over nonnegative number_t.
+  static number_t gcd_num(number_t a, number_t b) {
```
caballa: add assert that `a` cannot be negative?

caballa: It would be nice to move `gcd_num` and `fdiv` to bignums.hpp. 
Btw, I would call these functions `gcd` and `floor_div`

include/crab/domains/tvpi_dbm.hpp
```diff
@@ -0,0 +1,1967 @@
...
+      return v;
+    }
+    // The global template set is unsigned; anything outside its range is
+    // simply not in the template.
+    if (coefficient.fits_int64() && static_cast<int64_t>(coefficient) <=
+                                        std::numeric_limits<unsigned>::max()) {
+      const auto &coeffs = crab_domain_params_man::get().coefficients();
+      const unsigned c =
+          static_cast<unsigned>(static_cast<int64_t>(coefficient));
+      if (std::find(coeffs.begin(), coeffs.end(), c) != coeffs.end()) {
```
caballa: `std::find` is always linear regardless `coefficients` being sorted. You should use something like std::binary_search

include/crab/domains/tvpi_dbm.hpp
```diff
@@ -0,0 +1,1967 @@
...
+          return find_ghost_var(v, coeff);
+        });
+  }
+
+  /// Rewrite @p e scaled by @p coefficient: each @c ci*xi maps to
+  /// @c ghost(xi, ci*coefficient); @c none if such a ghost does not exist.
+  /// Used to propagate @c x := e into ghost space: @c ghost(x,c) :=
+  /// rewrite(e,c).
+  boost::optional<linear_expression_t>
+  try_rewrite_linear_expression(const linear_expression_t &e,
```
caballa: This is almost a duplicate of `rewrite_linear_expression`. Can we factorize code here?

include/crab/domains/tvpi_dbm.hpp
```diff
@@ -0,0 +1,1967 @@
...
+    }
+    res = res + (scale ? k / d : k);
+    return linear_constraint_t(res, cst.kind());
+  }
+
+  /// Mutating rewrite entry point (used by @c operator+=): creates the ghosts
+  /// it needs.  Returns @p cst unchanged when rewriting is impossible — the
+  /// @c ecst.equal(cst) signal (fixed mode, ghost outside the template set).
+  linear_constraint_t
+  rewrite_linear_constraint(const linear_constraint_t &cst) {
```
caballa: I would expect rewrite_linear_expression to be called as part of rewrite_linear_constraint

include/crab/domains/tvpi_dbm.hpp
```diff
@@ -0,0 +1,1967 @@
...
+   * @brief Add the ghost-space equality implied by the assignment @c x := e.
+   *
+   * Rewrites @p e term by term into ghost form (each @c ci*xi becomes
+   * @c ±ghost(xi,|ci|), the constant carries over) and adds @c x == rewrite(e)
+   * — e.g. @c x := 2*y + z  adds  @c x == ghost(y,2) + z, which the base
+   * domain stores as difference constraints.  If a required coefficient is
+   * outside the template the rewrite returns @p e unchanged and nothing is
+   * added.
+   */
+  void rewrite_assign(const variable_t &x, const linear_expression_t &e) {
```
caballa: Perhaps you do it in the caller but whenever you model assignment by addition of a constraint you need to `forget` the lhs of the assignment

include/crab/domains/tvpi_dbm.hpp
```diff
@@ -0,0 +1,1967 @@
...
+    if (gx == boost::none) {
+      return; // no ghost(x,c): nothing to maintain
+    }
+    auto gy = find_ghost_var(y, coefficient);
+    switch (op) {
+    case OP_MULTIPLICATION:
+    case OP_SDIV:
+    case OP_UDIV: // c*(y op z) == (c*y) op z for scalar z
+      if (gy != boost::none) {
+        m_absval.apply(op, *gx, *gy, z);
```
caballa, line 494: don't you need to scale `z` ?

include/crab/domains/tvpi_dbm.hpp
```diff
@@ -0,0 +1,1967 @@
...
+      *this = other;
+    } else if (other.is_bottom() || is_top()) {
+      // do nothing
+    } else {
+      CRAB_LOG("tvpi-dbm-join", crab::outs() << "[join]\n"
+                                             << *this << "\nwith\n"
+                                             << other << "\n");
+      m_absval |= other.m_absval;
+      m_vars.insert(other.m_vars.begin(), other.m_vars.end());
+      if constexpr (Params::normalize)
```
caballa: I would prefer just to call to `normalize` and then `normalize` does nothing if the flag is false

caballa (approved):
It's too much code for me to review carefully. I added some comments. Overall, it looks very impressive. I much prefer to merge and then fix bugs (if any) as we go.

Don't forget to rebase `dev` since I made many commits since you opened this PR

include/crab/domains/tvpi/ghost_variable_manager.hpp
LinerSu: Done. I moved both `marker` and `encode` into a private section, so `get_or_insert` is the only public entry point.

include/crab/domains/tvpi/ghost_variable_manager.hpp
LinerSu: Done, moved into the private section together with the marker above.

include/crab/domains/split_dbm.hpp
LinerSu: Done, and the first case turned out to be redundant as well: when both edges exist with weight 0, `edge_val(ai, bi)` already returns 0. The whole chain is now a single `elem` check plus a fallback return.

include/crab/domains/tvpi_dbm.hpp
LinerSu: The assert became unnecessary instead of being added. I moved `gcd_num` and `fdiv` into bignums.hpp per your other comment and implemented them with `mpz_gcd` and `mpz_fdiv_q`. `mpz_gcd` takes absolute values internally, so the nonnegativity precondition no longer exists.

LinerSu: Done. I moved them to [bignums.hpp](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/numbers/bignums.hpp#L122-L134) as `gcd` and `floor_div`, implemented with `mpz_gcd` and `mpz_fdiv_q`, and declared them as friends of `z_number` since `get_mpz_t()` is marked for removal. As a bonus, `floor_div` now handles negative divisors correctly, which the old truncate then correct version did not.

include/crab/domains/tvpi_dbm.hpp
LinerSu: Done. Sortedness is safe to rely on because the `fixed_tvpi_domain_params` constructor sorts the template, and that is documented as an invariant on the mutable accessor.

include/crab/domains/tvpi_dbm.hpp
LinerSu: Done. All rewriting entry points now share one primitive, [rewrite_terms](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/domains/tvpi_dbm.hpp#L303-L338), which maps each term `ci*xi` to `ghost(xi, |ci|*num/den)`. The plain ghost form is the `(1,1)` instance, the scaled form used by ghost updates is the `(c,1)` instance, and the Scaling rule for constraints is the `(1,d)` instance.

include/crab/domains/tvpi_dbm.hpp
LinerSu: They now share the same primitive, see the thread on line 395, but the delegation cannot go in the direction you suggested. The constraint path divides through by the gcd `d`, and that division is only meaning preserving under the relational operator. For example, over the integers `2x <= 5` becomes `x <= 2` by flooring the constant, which is valid as a constraint transformation but not as an expression transformation, since `2x` and `x` denote different values.

include/crab/domains/tvpi_dbm.hpp
LinerSu: It is done in the caller, as you guessed. `assign_impl` calls `m_absval.assign(x, e)` first, which kills `x`, and only then adds the ghost equality as a refinement of a value the base domain already computed. I renamed the function to `refine_assign_with_ghosts` to make that role clear and documented the ordering it depends on. Your comment also led me to a real bug in the self referential case: for `x := x + 2*y` the rewritten equality was `x == x + ghost(y,2)`, where the two `x` terms cancel and the meet degenerates to `ghost(y,2) == 0`, collapsing the state to bottom. The refinement now skips any rewritten rhs that still mentions `x`, see [the guard](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/domains/tvpi_dbm.hpp#L426-L438), and there is a regression test for it.

include/crab/domains/tvpi_dbm.hpp
LinerSu: For multiplication no scaling is needed: `c*(y*z)` equals `(c*y)*z`, so the multiplier passes through unchanged, and only addition and subtraction scale the operand. For division your question uncovered a real soundness bug. Truncated division does not commute with scaling: with `c = 3, z = 2, y = 7` we get `c*(y/z) = 9` but `(c*y)/z = 10`. I replaced that branch with a sound band derived from the division identity `y = sign(z)*|z|*x + r` with `|r| <= |z| - 1`, see [the new case](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/domains/tvpi_dbm.hpp#L502-L555). For example, `x := y/2` with `y` in `[10, 21]` now yields `2x <= y` and `y - 2x <= 1` instead of the false equality `2x == y`, which had wrongly implied that `y` is even. Every branch of the new code has a regression test in unittests-tvpi4.

include/crab/domains/tvpi_dbm.hpp
LinerSu: I kept `normalize()` itself unconditional because the lazy variant relies on clients calling it explicitly, so it must never become a no op. The boilerplate is gone though: a private `maybe_normalize()` holds the flag check and every call site uses it.

LinerSu:
Hi @caballa, thanks for the review. All comments are addressed, each answered in its own thread. Beyond them I did a full pass over the transfer functions, which found a few real soundness bugs, now fixed and tested. Five new commits on top of the reviewed four. The rebase onto `dev` comes after, so these changes stay reviewable against the version you commented on.

LinerSu:
**Soundness fixes.** The pass looked for steps that read a dimension an earlier step of the same transfer function had already changed. Each fix has a test that failed before it:

* `x := x + 2*y` collapsed to bottom. The ghost equality `x == x + ghost(y,2)` cancels the two `x` terms and degenerates to `ghost(y,2) == 0`. Fixed by [a guard in the refinement](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/domains/tvpi_dbm.hpp#L426-L438).
* `x := x + z` with `z == 2` turned `x == 5` into `x == 9`. The singleton reduction ran after the base operation and applied it twice; [it now runs before](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/domains/tvpi_dbm.hpp#L1724-L1744) and delegates the whole operation once.
* Coefficient 1 in the template made `rename` and `expand` abort, since `ghost(v,1)` is `v` itself. The ghost loops now skip it.
* The incremental saturation hook could mix facts about the old and the new value of the assigned variable and derive bottom on reachable states. It is [deferred](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/domains/tvpi_dbm.hpp#L1639-L1646) until the ghosts are re-established.

**Division.** The ghost propagation asserted `a*x == sign(z)*y`, which is wrong for truncated division whenever the remainder is nonzero. It now uses [the band](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/include/crab/domains/tvpi_dbm.hpp#L502-L555) from `y = sign(z)*a*x + r` with `|r| <= a - 1`, one sided when the sign of `y` is entailed. For example, `x := y/2` with `y` in `[10, 21]` gives `2x <= y` and `y - 2x <= 1` instead of the false equality, and ghosts are refreshed from their definition `ghost(x,c) := c*x` instead of being dropped.

**Tests.** New file [unittests-tvpi4.cc](https://github.com/seahorn/crab/blob/093f3ed7f9e26d8eab42d86aad39db36897b6d9f/tests/domains/tvpi/unittests-tvpi4.cc) with 22 cases: a regression test per fix above, first time coverage for `weak_assign`, `select`, `expand`, `rename`, int conversion, bitwise, `forget` and the lattice constants, `to_disjunctive_linear_constraint_system` now implemented by delegation, and full branch coverage of the division band for both sdiv and udiv.

include/crab/numbers/bignums.hpp
```diff
@@ -121,8 +121,20 @@ class z_number {
 
   void write(crab::crab_os &o) const;
 
+  // greatest common divisor; the result is always non-negative, whatever
+  // the signs of the operands
+  friend z_number gcd(z_number x, z_number y);
```
caballa: why `friend` ?

caballa: I would expect `z_number gcd(z_number x);`  so the two operands are `this` and `x`

include/crab/numbers/bignums.hpp
```diff
@@ -121,8 +121,20 @@ class z_number {
...
+
+  // division rounding the quotient towards -infinity, unlike operator/
+  // which truncates towards zero
+  friend z_number floor_div(z_number x, z_number y);
```
caballa: same here

include/crab/numbers/bignums.hpp
```diff
@@ -121,8 +121,20 @@ class z_number {
...
+
 }; // class z_number
 
+z_number gcd(z_number x, z_number y);
```
caballa: Why do we need these two functions?

caballa (changes requested):

caballa (changes requested):
Let us write properly the tests.
1. Do not use current approach (print all invariants and checks to standard output and diff with a file that contains the expected results)
2. Do not use `assert` for checking properties

Use boost test for all of this. When you run a CFG, you run already the checker so you know whether the CRAB assertions should pass or not. Let's start by checking those. For the unit tests, you already use `assert` to check the things you want. Use instead, boost test library

include/crab/numbers/bignums.hpp
LinerSu: Done. They are members now: `z_number gcd(z_number x) const` and `z_number floor_div(z_number x) const`, so the two operands are `this` and `x`, matching the style of the existing arithmetic API. Call sites read `a.gcd(b)` and `c.floor_div(g)`.

LinerSu: The friend was there only because `get_mpz_t()` is marked "to be removed" and I did not want a free function reaching through it. Moot now: both are plain member functions, no friend needed.

include/crab/numbers/bignums.hpp
LinerSu: Done, same change: `floor_div` is a member as well.

include/crab/numbers/bignums.hpp
LinerSu: Those were the namespace scope declarations of the two friend functions. Gone now that both are members.

LinerSu:
Done. All tvpi tests are ported to Boost.Test and merged into two files, nothing is printed to standard output, and the tvpi entries are removed from expected_results.out again since silent tests need no golden output.

[tvpi_cfg.cc](https://github.com/seahorn/crab/blob/d5ddb5b0ec9b4a5a556a9e45f5b07cce65f79194/tests/domains/tvpi/tvpi_cfg.cc) merges the two CFG files. Each assertion carries a debug_info id; the tests run the analyzer plus the assertion checker and check the resulting checks_db, both the per program totals and the verdict of every individual assertion.

[unittests-tvpi.cc](https://github.com/seahorn/crab/blob/d5ddb5b0ec9b4a5a556a9e45f5b07cce65f79194/tests/domains/tvpi/unittests-tvpi.cc) merges the six unit test files into per topic suites (transfer functions, incremental saturation, normalization, reduction rules, review regressions, division). Fixtures scope the coefficient template per suite. Every check went through BOOST_TEST; you were right about assert, those checks were dead under NDEBUG, and it was reassuring that all of them still pass once they actually run.

One practical note: the Boost log goes to stderr (both binaries pass --log_sink=stderr to themselves), so run_tests.sh sees empty stdout and failures stay visible in ctest via the exit code.

caballa:
Merged already. Thank you @LinerSu !

PR #68, opened 2026-08-15
Add a shared container-printing facility (crab::seq / crab::kv / crab::opt)

LinerSu:
## Motivation

crab has no shared way to print an STL container to a `crab_os`, so every
domain and analysis hand-rolls its own separator loop. A sweep of `include/` +
`lib/` found ~105 such loops across ~50 `write`/`dump`/`operator<<` functions,
using three different separator idioms and five disagreeing bracket
conventions, plus 8 named ad-hoc helpers (three of them byte-identical copies
of the same vector printer). Separately, `boost::optional` — used ~380 times in
`include/` — cannot be printed at all today: `o << opt` does not compile, so
every debug print of one is a manual `if (opt) ... else` at the call site.

## What this adds

One self-contained C++14 header, `include/crab/support/print.hpp` (no `.cpp`,
no change to `os.hpp` or any build file):

```cpp
o << crab::seq(v);                        // [a, b, c]
o << crab::seq(v, print::fmt_set());      // {a, b, c}
o << crab::seq(v).sep("; ").bare();       // a; b; c
o << crab::kv(m);                         // {k1 -> v1; k2 -> v2}
o << crab::opt(x);                        // *x, or "none" when empty
print::print_range_with(o, r, fn, fmt);   // fn(crab_os&, const Elem&)
print::separator sep(o, ", ");            // for loops that filter elements or
for (...) { ...; sep.next() << e; }       // share a "first" flag across loops
print::to_string(x);                      // anything -> std::string
```

Elements dispatch automatically through a small SFINAE ladder:
`operator<< > std::pair > container (recurse) > optional-like > write(crab_os&)`,
with a readable `static_assert` if none applies. `crab::opt` covers
`boost::optional`, `shared_ptr`, `unique_ptr` (and `std::optional` under a
future C++17 build) structurally, without including any of their headers.

## Design decisions

- **No generic `operator<<(crab_os&, const T&)`.** `crab_os`'s 13 stream
  overloads are non-template virtual members that `bool`, `float`, `short`,
  enums and `T*` reach only by promotion/conversion; a `const T&` template
  would be an identity match and silently steal them. The wrappers' hidden
  friend `operator<<` can only ever match the wrapper.
- **`operator<<` is the top dispatch rank**, so an element that could already
  be streamed prints byte-identically to before, by construction. This
  matters because several crab types (e.g. `linear_constraint_t`) have both
  iterators and an `operator<<`.
- **Raw pointers are never auto-dereferenced.** `o << p` keeps printing an
  address; the pointee requires an explicit `crab::opt(p)`. This is what makes
  the optional-like rank unable to change any existing output.
- `std::string` is excluded from container detection (`vector<string>` prints
  `[hi, there]`, not `[[h,i],...]`), and detection uses member `begin()/end()`
  only, so string literals are not ranges.
- C++14 throughout (`make_void` struct per CWG 1558, `priority_tag` ladder,
  no `if constexpr`).

## Changes to existing code

1. `___print___` (debug.hpp) now takes `const ArgTypes &...` instead of
   by-value — output-neutral, and stops deep-copying every `std::string`
   argument at the ~590 `CRAB_ERROR`/`CRAB_WARN` call sites.
2. 26 hand-rolled printing loops that live inside `CRAB_LOG` blocks (or in
   helpers reachable only from them) are converted to the facility, across 20
   files. These sites are invisible to the golden tests (which discard stderr
   and never enable logging), making them the safest possible first users.
   Two debug-only quirks get fixed on the way, both called out here
   deliberately:
   - the 7-copy `Renaming {a;b;} with c;d;}:` cluster (apron, elina,
     sparse_dbm, split_dbm, term_equiv, split_oct, herbrand) was missing the
     opening `{` of the second list in all seven copies; it now prints
     `Renaming {a;b;} with {c;d;}:`.
   - `offset_map_t::write` printed a stray `,` before `}` when the last cell
     of a set was removed; the `print::separator` rewrite fixes it.

   Remaining `CRAB_LOG` loops were left alone on purpose: per-element
   "\t...\n" dumps (a range call is no shorter), the split_oct `rev_map`
   loops (they need the index and a filter with a trailing separator), and
   one site inside dead `#if 0` code.

## Verification

- Golden suite (`tests/run_tests.sh`, default variant, 83 binaries):
  byte-identical before and after every commit.
- The debug output itself was verified, not assumed: stdout+stderr of 19 test
  binaries captured with the relevant `--log` tags before and after the
  conversion. The only difference across ~8200 lines is the intended
  missing-brace fix at the 7 Renaming sites.
- The header was probe-tested standalone (traits, every preset, nesting,
  `vector<string>`, optionals/smart pointers, raw-pointer non-deref,
  const/non-const `write()`, empty ranges) with clang++-14 and g++-10 at
  `-std=c++14 -Wall -Wextra`, both clean.
- apron/elina TUs cannot be compiled locally (no toolchain); their two
  converted hunks are pattern-identical to the five sibling copies that do
  compile and run here. CI with those backends is the real check, as are the
  elina/boxes golden variants.

cc @caballa -- curious what you think of the approach before any further conversion work.

### DIFF
--- a/include/crab/support/print.hpp
+++ b/include/crab/support/print.hpp
@@ -1,0 +1,577 @@
+#pragma once
+
+/*
+ * A small facility for printing containers (and optional-like values) to a
+ * crab_os, so that write()/operator<< implementations do not need to
+ * hand-roll separator loops.
+ *
+ * Quick reference (o is a crab_os):
+ *
+ *   o << crab::seq(v);                       // [a, b, c]
+ *   o << crab::seq(v, print::fmt_set());     // {a, b, c}
+ *   o << crab::seq(v).sep("; ").bare();      // a; b; c
+ *   o << crab::kv(m);                        // {k1 -> v1; k2 -> v2}
+ *   o << crab::opt(x);                       // value of *x, or "none"
+ *   o << crab::opt(x, "<absent>");           // custom empty marker
+ *
+ *   print::print_range(o, v, print::fmt_seq_tight());   // [a,b,c]
+ *   print::print_range_with(o, v, fn);       // fn(crab_os&, const elem&)
+ *   print::to_string(x);                     // print anything to a std::string
+ *
+ *   print::separator sep(o, ", ");           // for loops that filter elements
+ *   for (auto &e : v) {                      // or span several sibling loops
+ *     if (skip(e)) continue;
+ *     sep.next() << e;
+ *   }
+ *
+ * Elements are printed by trying, in order: std::pair (first -> second),
+ * container (recurse), optional-like (dereference or print the empty marker),
+ * operator<<, then write(crab_os&). See print_element below for the
+ * rationale of that order.
+ *
+ * Design notes:
+ *
+ * - There is deliberately NO generic operator<<(crab_os&, const T&).
+ *   crab_os's operator<< overloads are non-template members, and bool, float,
+ *   short, enums and T* reach them only by promotion/conversion; a generic
+ *   const T& template would be an identity match and silently steal them.
+ *   The seq/kv/opt wrappers sidestep this: their operator<< is a plain
+ *   non-template hidden friend that can only ever match the wrapper.
+ *
+ * - Raw pointers are NEVER dereferenced automatically. A raw T* keeps
+ *   binding to crab_os::operator<<(const void*) and prints as an address,
+ *   as it always has. If a caller wants the pointee they must say so
+ *   explicitly with crab::opt(p).
+ *
+ * - This header is C++14: no if constexpr, no std::void_t, no fold
+ *   expressions, hence the make_void / priority_tag machinery below.
+ */
+
+#include <crab/support/os.hpp>
+
+#include <string>
+#include <type_traits>
+#include <utility>
+
+namespace crab {
+namespace print {
+
+/*
+ * Describes ONE level of container printing: brackets, separator, the arrow
+ * between the members of a pair, what to print for an empty container, the
+ * marker for an empty optional-like element, and whether the separator also
+ * trails the last element (the "{a;b;c;}" debug style).
+ *
+ * Nested elements do not inherit the enclosing format: a nested container is
+ * printed with its own type default (fmt_map() if its elements are pairs,
+ * fmt_seq() otherwise). So map<K, vector<V>> prints {k -> [a, b], ...}.
+ *
+ * The setters return a modified copy, so formats chain:
+ *   format().brackets("{", "}").sep("; ")
+ */
+class format {
+public:
+  const char *m_open = "[";
+  const char *m_close = "]";
+  const char *m_sep = ", ";
+  const char *m_arrow = " -> ";
+  // printed instead of m_open/m_close when the range is empty (nullptr means
+  // print m_open followed by m_close)
+  const char *m_empty = nullptr;
+  // marker for an empty optional-like element. "none" matches the existing
+  // convention (e.g. cfg.hpp); it is not "_|_", which in crab means bottom,
+  // a lattice value -- an absent value is not a bottom value.
+  const char *m_none = "none";
+  bool m_trailing = false;
+
+  format sep(const char *s) const {
+    format f(*this);
+    f.m_sep = s;
+    return f;
+  }
+  format brackets(const char *open, const char *close) const {
+    format f(*this);
+    f.m_open = open;
+    f.m_close = close;
+    return f;
+  }
+  format bare() const { return brackets("", ""); }
+  format arrow(const char *a) const {
+    format f(*this);
+    f.m_arrow = a;
+    return f;
+  }
+  format empty(const char *e) const {
+    format f(*this);
+    f.m_empty = e;
+    return f;
+  }
+  format none(const char *n) const {
+    format f(*this);
+    f.m_none = n;
+    return f;
+  }
+  format trailing(bool b = true) const {
+    format f(*this);
+    f.m_trailing = b;
+    return f;
+  }
+};
+
+/*
+ * Presets, named after the conventions they reproduce so that migrating an
+ * existing printer is a lookup rather than a judgement call.
+ */
+// [a, b, c]
+inline format fmt_seq() { return format(); }
+// [a,b,c]
+inline format fmt_seq_tight() { return format().sep(","); }
+// {a, b, c}
+inline format fmt_set() { return format().brackets("{", "}"); }
+// {a,b,c}
+inline format fmt_set_tight() { return fmt_set().sep(","); }
+// {k1 -> v1; k2 -> v2} -- the patricia-tree map printers
+inline format fmt_map() { return format().brackets("{", "}").sep("; "); }
+// same bytes as fmt_map(); named separately so patricia call sites document
+// which convention they came from
+inline format fmt_patricia() { return fmt_map(); }
+// [|v1, v2|] -- the graph printers
+inline format fmt_graph() { return format().brackets("[|", "|]"); }
+// {a;b;c;} -- the CRAB_LOG debug loops
+inline format fmt_debug() {
+  return format().brackets("{", "}").sep(";").trailing();
+}
+
+/*
+ * A stateful joiner for the loops that cannot become a single range call:
+ * loops that `continue`-filter elements, or two sibling loops sharing one
+ * "first" flag (e.g. the split_dbm printer). The first call to next() prints
+ * nothing; every later call prints the separator. Either way it returns the
+ * stream, so the natural use is:
+ *
+ *   separator sep(o, ", ");
+ *   for (...) { if (filtered) continue; sep.next() << elem; }
+ */
+class separator {
+  crab_os &m_o;
+  const char *m_sep;
+  bool m_first;
+
+public:
+  separator(crab_os &o, const char *sep = ", ")
+      : m_o(o), m_sep(sep), m_first(true) {}
+  crab_os &next() {
+    if (!m_first) {
+      m_o << m_sep;
+    }
+    m_first = false;
+    return m_o;
+  }
+  // true iff next() has not fired yet -- for "print a footer only if we
+  // printed anything" call sites
+  bool first() const { return m_first; }
+  void reset() { m_first = true; }
+};
+
+template <typename Range>
+void print_range(crab_os &o, const Range &r, const format &fmt);
+
+namespace detail {
+
+// C++14 has no std::void_t. Hand-rolled through a struct rather than a
+// direct alias: with a direct alias some compilers do not guarantee that
+// unused alias arguments trigger SFINAE (CWG 1558).
+template <typename...> struct make_void { using type = void; };
+template <typename... Ts> using void_t = typename make_void<Ts...>::type;
+
+// Rank ordering for the dispatch ladder below: an argument of type
+// priority_tag<N> prefers an overload taking priority_tag<N> over one taking
+// priority_tag<N-1> (derived-to-base conversion), so ranks are tried
+// highest-first and SFINAE'd-out ranks fall through.
+template <unsigned N> struct priority_tag : priority_tag<N - 1> {};
+template <> struct priority_tag<0> {};
+
+template <typename T> struct always_false : std::false_type {};
+
+template <typename T> struct is_pair : std::false_type {};
+template <typename A, typename B>
+struct is_pair<std::pair<A, B>> : std::true_type {};
+
+// Ranges are detected through MEMBER begin()/end() only, never std::begin:
+// with std::begin every string literal (const char[N]) would count as a
+// range. std::string is excluded explicitly, and the reason is nesting, not
+// ambiguity: it has member begin()/end(), so without the exclusion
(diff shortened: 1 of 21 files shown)


include/crab/support/print.hpp
```diff
@@ -0,0 +1,594 @@
...
+    return seq_ref(*m_r, m_fmt.trailing(b));
+  }
+  void write(crab_os &o) const { print::print_range(o, *m_r, m_fmt); }
+  friend crab_os &operator<<(crab_os &o, const seq_ref &s) {
+    s.write(o);
+    return o;
+  }
+};
+
+template <typename Range> seq_ref<Range> seq(const Range &r) {
```
caballa, line 553: I would call either `print_seq` or move `seq` to the `print` namespace. We need to think well about ergonomics so it's easier for the dev to remember. Same for all the functions below.

include/crab/support/print.hpp
```diff
@@ -0,0 +1,594 @@
...
+  return seq_ref<Range>(r, print::fmt_seq());
+}
+template <typename Range>
+seq_ref<Range> seq(const Range &r, const print::format &fmt) {
```
caballa: same thing, `print_seq`

include/crab/support/print.hpp
```diff
@@ -0,0 +1,594 @@
...
+  return seq_ref<Range>(r, fmt);
+}
+
+// A range of pairs printed as a map: o << crab::kv(m) -> {k1 -> v1; k2 -> v2}
+template <typename Map> class kv_ref {
```
caballa, line 562: `print_kv_ref`

include/crab/support/print.hpp
```diff
@@ -0,0 +1,594 @@
...
+  kv_ref arrow(const char *a) const { return kv_ref(*m_m, m_fmt.arrow(a)); }
+  kv_ref empty(const char *e) const { return kv_ref(*m_m, m_fmt.empty(e)); }
+  void write(crab_os &o) const { print::print_range(o, *m_m, m_fmt); }
+  friend crab_os &operator<<(crab_os &o, const kv_ref &m) {
+    m.write(o);
+    return o;
+  }
+};
+
+template <typename Map> kv_ref<Map> kv(const Map &m) {
```
caballa, line 582: `print_kv`

include/crab/support/print.hpp
```diff
@@ -0,0 +1,594 @@
...
+  return kv_ref<Map>(m, print::fmt_map());
+}
+template <typename Map> kv_ref<Map> kv(const Map &m, const print::format &fmt) {
```
caballa: `print_kv`

include/crab/support/print.hpp
```diff
@@ -0,0 +1,594 @@
...
+      o << m_none;
+    }
+  }
+  friend crab_os &operator<<(crab_os &o, const opt_ref &x) {
+    x.write(o);
+    return o;
+  }
+};
+
+template <typename T> opt_ref<T> opt(const T &x, const char *none = "none") {
```
caballa, line 615: `print_opt`

include/crab/support/print.hpp
```diff
@@ -0,0 +1,594 @@
...
+ * the wrapped object, so use them directly in a print expression rather
+ * than storing them.
+ *
+ * operator<< is a hidden friend: a non-template function that only ADL can
+ * find and that can only ever match the wrapper, so it cannot interfere
+ * with crab_os's own non-template operator<< overloads.
+ */
+
+// A range printed as a sequence: o << crab::seq(v) -> [a, b, c]
+template <typename Range> class seq_ref {
```
caballa, line 531: What is the difference between print_range and seq_ref?

caballa (changes requested):
It's really nice. I still need to look more careful but I wrote already some comments.

include/crab/support/print.hpp
LinerSu: Done in 325a960a: moved `seq`/`kv`/`opt` into `crab::print`, so everything print-related is in one namespace. Call sites inside crab now read `print::seq(v, print::fmt_debug())`. I kept the short names since the namespace already gives the context.

include/crab/support/print.hpp
LinerSu: Done in 325a960a (moved into `crab::print`).

include/crab/support/print.hpp
LinerSu: Done in 325a960a (moved into `crab::print`).

include/crab/support/print.hpp
LinerSu: Done in 325a960a (moved into `crab::print`).

include/crab/support/print.hpp
LinerSu: Done in 325a960a (moved into `crab::print`).

include/crab/support/print.hpp
LinerSu: Done in 325a960a (moved into `crab::print`).

include/crab/support/print.hpp
LinerSu: `seq_ref` is just the object that `print::seq(v)` returns; its `write()` calls `print_range`. Same output, different call styles:

```cpp
// wrapper: a value you can put in a << chain or pass to CRAB_ERROR
o << "vars=" << print::seq(vs) << "\n";
CRAB_ERROR("cannot unify ", print::seq(params));

// function: prints right now, inside a bigger write()
print::print_range(o, vs, print::fmt_set());

// function with a callback, when you control how each element prints
// (e.g. patricia trees, whose binding is not a std::pair)
print::print_range_with(o, m_env,
    [](crab_os &o, const auto &p) { o << p.first << " -> " << p.second; },
    print::fmt_patricia());
```

The wrapper cannot do the callback form, and the functions cannot sit in a `<<` chain. That is the only reason both exist. Documented now in the header (325a960a).

LinerSu:
Also drafted a developer how-to for the wiki, organized by data structure (vectors/sets, maps, patricia trees, optionals, plus a full `write()` example for a new domain). Content below, collapsed. Feel free to drop it into the wiki wherever it fits, and happy to adjust.

<details>
<summary><b>HowToPrintProgramStates.md</b> (click to expand)</summary>

Every abstract domain implements `void write(crab_os&)`, and most analyses
print containers in `CRAB_LOG` messages. The header
`crab/support/print.hpp` provides one shared way to do this, so you do not
need to hand-roll a separator loop (the library used to contain ~105 of
them, in five disagreeing styles).

This page is organized by **what you have in front of you**, not by API.

```cpp
#include <crab/support/print.hpp>
```

Code inside `namespace crab` writes `print::seq(v)`. Code outside (clam,
seahorn) writes `crab::print::seq(v)`, or once per file:

```cpp
namespace cp = crab::print;
```

# TL;DR

| You have                                   | You write                                         | It prints              |
|--------------------------------------------|---------------------------------------------------|------------------------|
| `std::vector<variable_t> vs`               | `o << print::seq(vs)`                             | `[x, y, z]`            |
| `std::set<variable_t> s`                   | `o << print::seq(s, print::fmt_set())`            | `{x, y, z}`            |
| `std::map` / `std::unordered_map` `m`      | `o << print::kv(m)`                               | `{x -> [0,5]; y -> 1}` |
| patricia tree / `ikos::separate_domain`    | `print_range_with` + callback (see below)         | `{x -> [0,5]; ...}`    |
| `boost::optional<T> x`                     | `o << print::opt(x)`                              | value, or `none`       |
| `shared_ptr` / `unique_ptr`                | `o << print::opt(p)`                              | pointee, or `none`     |
| a loop that filters elements               | `print::separator` (see below)                    | —                      |
| anything, into a `std::string`             | `print::to_string(x)`                             | —                      |

Elements print themselves automatically: anything with an
`operator<<(crab_os&, ...)` or a `write(crab_os&)` method works, including
nested containers (`map<K, vector<V>>` prints `{k -> [a, b]; ...}`) and
`vector<boost::optional<T>>` (`[a, none, c]`).

# Sequences and sets

```cpp
std::vector<variable_t> vs = ...;
o << print::seq(vs);                          // [x, y, z]
o << print::seq(vs, print::fmt_set_tight());  // {x,y,z}
o << print::seq(vs).sep("; ").bare();         // x; y; z
```

The presets (`fmt_seq`, `fmt_seq_tight`, `fmt_set`, `fmt_set_tight`,
`fmt_map`, `fmt_patricia`, `fmt_graph`, `fmt_debug`) are named after the
formats that already exist in the library, so converting an old printer is
a lookup, not a judgement call. `fmt_debug()` is the trailing-separator
`{a;b;c;}` style used in most `CRAB_LOG` messages.

# Maps

```cpp
std::unordered_map<variable_t, variable_t> m = ...;
o << print::kv(m);                            // {x -> y; z -> w}
o << print::kv(m, print::fmt_set());          // {x -> y, z -> w}
o << print::kv(m).arrow(" = ").sep(", ");     // {x = y, z = w}
```

**Determinism warning:** `unordered_map` iteration order is unspecified. If
the output must be stable (test expectations, diffs between runs), collect
and sort the keys first, then print with a lookup callback:

```cpp
std::vector<variable_t> keys;
keys.reserve(m.size());
for (auto &kv : m) { keys.push_back(kv.first); }
std::sort(keys.begin(), keys.end());
print::print_range_with(o, keys,
    [&m](crab_os &o, const variable_t &k) { o << k << " -> " << m.at(k); },
    print::fmt_map());
```

(`numerical_packing.hpp` does exactly this.)

# Patricia trees (`ikos::separate_domain`, environments)

Patricia-tree iterators yield a binding that has `first`/`second` members
but is **not** a `std::pair`, so `print::kv` refuses it at compile time.
Use the callback form — this is deliberate, since the `Key`/`Value` types
are often client-supplied and the callback pins down exactly how they
print:

```cpp
// m_env : ikos::separate_domain<variable_t, interval_t>
print::print_range_with(o, m_env,
    [](crab_os &o, const auto &p) { o << p.first << " -> " << p.second; },
    print::fmt_patricia());                   // {x -> [0,5]; y -> [1,2]}
```

# Optionals, smart pointers, raw pointers

```cpp
boost::optional<invariant_t> inv = ...;
o << print::opt(inv);                         // the value, or "none"
o << print::opt(inv, "<not computed>");       // custom empty marker
```

`o << inv` does not compile (and never did) — `print::opt` is the way.
Raw pointers keep printing as **addresses** everywhere, exactly as before;
if you want the pointee you must say so explicitly with `print::opt(p)`.
The empty marker defaults to `"none"`, not `_|_` — an absent value is not
a bottom value.

# Loops that cannot become one call

If the loop filters elements, or two sibling loops share one comma state,
keep the loop and replace only the separator bookkeeping:

```cpp
// before                                     // after
bool first = true;                            print::separator sep(o, ", ");
for (auto &e : elems) {                       for (auto &e : elems) {
  if (skip(e)) continue;                        if (skip(e)) continue;
  if (!first) o << ", ";                        sep.next() << e;
  first = false;                              }
  o << e;
}
```

`sep.next()` prints nothing the first time and the separator every time
after, and returns the stream. One `separator` can span several loops.
`sep.first()` tells you whether anything was printed (for conditional
footers).

# A complete `write()` for a new domain

Substates compose; each picks the recipe for its own data structure:

```cpp
void write(crab_os &o) const override {
  if (is_bottom()) { o << "_|_"; return; }
  if (is_top())    { o << "top"; return; }
  // m_env:      ikos::separate_domain<variable_t, interval_t>  (patricia)
  // m_ghosts:   std::unordered_map<variable_t, variable_t>
  // m_worklist: std::vector<variable_t>
  // m_csts:     boost::optional<linear_constraint_system_t>
  o << "(env=";
  print::print_range_with(o, m_env,
      [](crab_os &o, const auto &p) { o << p.first << " -> " << p.second; },
      print::fmt_patricia());
  o << ", ghosts=" << print::kv(m_ghosts, print::fmt_set())
    << ", wl="     << print::seq(m_worklist)
    << ", csts="   << print::opt(m_csts, "<not computed>") << ")";
}
```

The wrappers also work as `CRAB_ERROR` / `CRAB_WARN` / `CRAB_LOG`
arguments, and `print::to_string(x)` prints anything printable into a
`std::string` (replacing the `crab_string_os` dance).

# When the compiler refuses

If an element type is not printable you get one readable error:

    static_assert failed ... "crab::print: this element type is not
    printable. It needs an operator<<(crab_os&, const T&) or a
    write(crab_os&) method; alternatively use print_range_with with an
    explicit callback."

Those are the two ways out: give the type a printer, or print it through a
callback lambda.

# What it is *not* for

- **Sign-aware arithmetic formatting** (`linear_expression::write` printing
  `x - 2*y + 1`): that is formatting logic, not separator logic.
- **Per-element prefixed dumps** (`"\t" << elem << "\n"` per line): a
  callback can do it, but it is usually no shorter than the loop — use your
  judgement.
- **Changing an existing observable format**: the golden tests
  (`tests/expected_results*.out`) lock printed output. The facility can
  reproduce every existing convention byte-for-byte via the presets — when
  converting an old printer, pick the preset that matches and verify
  against the goldens.

</details>

Issue #71, opened 2026-08-19
API: functional project/forget (make_projection/make_forget) to avoid copy-then-project


LinerSu:
`project` and `forget` are in-place in `abstract_domain_api`, so every
caller needing a *restricted view* of a state has to copy the whole domain
first and then project the copy:

```cpp
Dom copy(dom);
copy.project(vars);   // most of the copied dimensions are thrown away
```

This pattern shows up across crab, independent of any particular domain:

- generic analyzer: `print_invariants` handling in
  `fwd_abs_transformers.hpp` (`abs_dom_t copy(m_inv); copy.project(vars);`)
  — paid by whatever domain the analysis runs; for `split_dbm`/`split_oct`
  the copy duplicates the entire difference-bound graph and its closure
  state,
- `region/ghost_variable_manager.hpp` (two sites, used by `region_domain`
  and `object_domain`),
- `object_domain`'s cache/base reduction (one site carries a FIXME asking
  for exactly this: "avoid copying, use efficient project onto").

Meanwhile some backends already have cheaper functional forms that the
in-place API cannot expose: Apron/ELINA take a `destructive` flag and crab
already calls the non-destructive form internally
(`ap_abstract0_forget_array(man, /*destructive=*/false, ...)`) — the extra
copy is pure overhead on top of it. Graph domains could build the
restricted sub-graph directly; environment domains a filtered tree.

**Proposal**: keep `project`/`forget` unchanged and add two pure virtuals
to `abstract_domain_api`, named after the existing in-place/functional
precedent in the API (`set_to_top` vs `make_top`):

```cpp
// Return a new abstract state: *this restricted to variables
virtual Dom make_projection(const variable_vector_t &variables) const = 0;
// Return a new abstract state: *this with variables forgotten
virtual Dom make_forget(const variable_vector_t &variables) const = 0;
```

with the reusable defaults provided the same way the existing defaults
are — opt-in macros in `abstract_domain_macros.def` (one line per domain,
like `DEFAULT_SELECT` / `DEFAULT_WEAK_ASSIGN`), keeping visible which
domains could implement the operation more efficiently:

```cpp
#define DEFAULT_MAKE_PROJECTION(DOM)                                     \
  virtual DOM make_projection(const variable_vector_t &vs)               \
      const override {                                                   \
    DOM res(*this);                                                      \
    res.project(vs);                                                     \
    return res;                                                          \
  }
// DEFAULT_MAKE_FORGET(DOM): same shape with forget
```

[...]

PR #72, opened 2026-08-20
Functional project/forget: make_projection and make_forget (#71)

LinerSu:
Implements #71: functional counterparts of the in-place `project`/`forget`,
so a caller that needs a restricted view of a state no longer has to copy
the whole domain and project the copy. The in-place operations are
unchanged. Names follow the `set_to_top`/`make_top` precedent.

## What is in the PR (4 commits)

1. **API + reusable defaults** — two pure virtuals in `abstract_domain_api`:

   ```cpp
   virtual Dom make_forget(const variable_vector_t &) const = 0;
   virtual Dom make_projection(const variable_vector_t &) const = 0;
   ```

   Defaults (copy + in-place op) are opt-in macros in
   `abstract_domain_macros.def` (`DEFAULT_MAKE_PROJECTION` /
   `DEFAULT_MAKE_FORGET`), one line per domain, per the file's convention
   of keeping visible which domains could do better. All implementers
   adopt the defaults in this commit; the generic type-erasure wrappers
   forward through their four layers like `make_top`;
   `dummy_abstract_domain` aborts like its other methods. Zero behavior
   change.

2. **Specialized implementations** where the copy is genuinely expensive:
   - `split_dbm`: one closure pass, then materialize only the induced
     subgraph over the kept vertices (`GraphPerm` view + `graph_t::copy`
     through the private state constructor). Closed distances already
     carry every path through dropped vertices, so the subgraph needs no
     re-normalization. `make_forget` projects onto the tracked complement.
   - `split_oct`: same recipe; a (v+, v-) vertex pair per variable and no
     zero vertex, following the domain's own widening prologue.
   - `apron`/`elina`: the libraries are functional underneath —
     `*_forget_array(destructive=false)` already returns a fresh state
     (the in-place `forget` relies on the same contract); wrap that state
     directly instead of duplicating the whole library value first.
   - reduced products, `numerical_congruence`, `powerset`:
     component/element-wise forwarding so the components' specializations
     apply; `flat_boolean_numerical` forwards `make_projection` through
     the product (its side environments start top, exactly what the
     in-place project does to them).

   Deliberately NOT specialized, with rationale in the code: the
   `separate_domain`-based leaf domains (their copy is an O(1)
   patricia-root share, so the default is already optimal),
   `array_smashing` (the array-var to scalar-ghost translation would be
   duplicated), and the remaining wrappers.

3. **Caller migrations** — the pure copy-then-project sites in-tree:
   `ghost_variable_manager` (two sites inside region_domain's join/meet
   plumbing), the `print_invariants` intrinsic, and the top-down
   analyzer's summary precondition. A systematic sweep found no other
   pure sites; the remaining candidates mutate the copy in between and
   are listed in the commit message.

4. **Tests** — `tests/domains/unittests-make-projection.cc` (Boost.Test,
   silent on stdout, so no golden-file entries):
   - `functional_law`: for every domain in the matrix (intervals,
     constant, sign, numerical congruence, disjunctive intervals,
     split_dbm, split_oct, term, boolean+numerical, ELINA oct/pk when
     available), `make_projection`/`make_forget` equal the
     copy-then-in-place results by mutual inclusion, over a relational
     state, bottom, top, empty set, untracked variables, and
     all-tracked-variables.
   - `semantics`: a relation that only holds THROUGH a dropped variable
     survives a split_dbm projection; bounds survive split_oct's pair
     encoding; the functional forms leave the source state untouched.

   Beyond these, the functional forms are handy for writing tests in
   general: a restricted view of a state can be inspected without
   mutating the state under test.

## Validation

- The committed law tests above, plus the full test suite (ctest, golden
  sweeps) at every commit.
- Verified on BOTH library backends locally: the ELINA build (default
  config) and a separate `-DCRAB_USE_APRON=ON` build — full compile, the
  apron golden sweep, and the law harness over apron oct/pk/box
  (432 checks) all pass.

### DIFF
--- a/include/crab/domains/abstract_domain.hpp
+++ b/include/crab/domains/abstract_domain.hpp
@@ -395,6 +395,26 @@ public:
   // Project the abstract domain onto variables (dual to forget)
   virtual void project(const variable_vector_t &variables) = 0;
 
+  // Return a new abstract state: *this with variables forgotten.
+  // Functional counterpart of forget (set_to_top vs make_top precedent).
+  // A default (copy + in-place forget) is provided so that domains
+  // maintained OUTSIDE this repository (e.g. the PPLite native wrapper)
+  // keep compiling; in-tree domains opt in explicitly via the
+  // DEFAULT_MAKE_FORGET macro or provide a native implementation.
+  virtual Dom make_forget(const variable_vector_t &variables) const {
+    Dom res(static_cast<const Dom &>(*this));
+    res.forget(variables);
+    return res;
+  }
+
+  // Return a new abstract state: *this restricted to variables.
+  // Functional counterpart of project. Default as for make_forget.
+  virtual Dom make_projection(const variable_vector_t &variables) const {
+    Dom res(static_cast<const Dom &>(*this));
+    res.project(variables);
+    return res;
+  }
+
   // Make a new copy of var without relating var with new_var
   virtual void expand(const variable_t &var, const variable_t &new_var) = 0;
 
--- a/include/crab/domains/abstract_domain_macros.def
+++ b/include/crab/domains/abstract_domain_macros.def
@@ -297,6 +297,25 @@ virtual void weak_assign_bool_var(const variable_t &lhs, const variable_t &rhs, 
    }                                                                         \
 }
 
+/* Default implementations of make_projection/make_forget: copy the
+   whole state, then project/forget in place. Domains whose
+   representation allows building the restricted state directly (e.g.
+   environments, DBMs, apron/elina) should implement these natively. */
+#define DEFAULT_MAKE_PROJECTION(DOM)                                          \
+  virtual DOM make_projection(const variable_vector_t &vars)                  \
+      const override {                                                        \
+    DOM res(*this);                                                           \
+    res.project(vars);                                                        \
+    return res;                                                               \
+  }
+
+#define DEFAULT_MAKE_FORGET(DOM)                                              \
+  virtual DOM make_forget(const variable_vector_t &vars) const override {     \
+    DOM res(*this);                                                           \
+    res.forget(vars);                                                         \
+    return res;                                                               \
+  }
+
 #define DEFAULT_ENTAILS(DOM)						\
 virtual bool entails(const linear_constraint_t &cst) const override {   \
  if (is_bottom()) {							\
--- a/include/crab/domains/generic_abstract_domain.hpp
+++ b/include/crab/domains/generic_abstract_domain.hpp
@@ -27,6 +27,10 @@ namespace domains {
 
 template <typename Variable>
 class abstract_domain final : abstract_domain_api<abstract_domain<Variable>> {
+  // the base is private: let its defaulted make_forget/make_projection
+  // downcast to this class
+  friend class abstract_domain_api<abstract_domain<Variable>>;
+
 public:
   using number_t = typename Variable::number_t;
   using varname_t = typename Variable::varname_t;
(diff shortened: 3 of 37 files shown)


LinerSu:
The pplite-native CI failure is fixed: the PPLite native wrapper implements `abstract_domain_api` from the PPLite tree, so pure virtuals broke it. The two methods now carry a default implementation (copy + in-place op) in the base class for out-of-tree compatibility; in-tree domains keep the explicit `DEFAULT_MAKE_*` macro opt-ins. Force-pushed; no other changes.

priyasiddharth:
Same comment as https://github.com/seahorn/crab/pull/75

include/crab/domains/generic_abstract_domain.hpp
```diff
@@ -645,6 +653,18 @@ class abstract_domain final : abstract_domain_api<abstract_domain<Variable>> {
...
     }
+    std::unique_ptr<abstract_domain_concept>
+    make_forget(const variable_vector_t &variables) const override {
+      std::unique_ptr<abstract_domain_concept> res(
+          new abstract_domain_model(m_inv.make_forget(variables)));
+      return res;
+    }
+    std::unique_ptr<abstract_domain_concept>
+    make_projection(const variable_vector_t &variables) const override {
+      std::unique_ptr<abstract_domain_concept> res(
```
caballa, line 664: use make_unique

include/crab/domains/generic_abstract_domain.hpp
```diff
@@ -645,6 +653,18 @@ class abstract_domain final : abstract_domain_api<abstract_domain<Variable>> {
     void project(const variable_vector_t &variables) override {
       m_inv.project(variables);
     }
+    std::unique_ptr<abstract_domain_concept>
+    make_forget(const variable_vector_t &variables) const override {
+      std::unique_ptr<abstract_domain_concept> res(
```
caballa, line 658: use make_unique, crab uses now c++14

include/crab/domains/abstract_domain.hpp
```diff
@@ -395,6 +395,26 @@ template <class Dom> class abstract_domain_api:
   // Project the abstract domain onto variables (dual to forget)
   virtual void project(const variable_vector_t &variables) = 0;
 
+  // Return a new abstract state: *this with variables forgotten.
+  // Functional counterpart of forget (set_to_top vs make_top precedent).
+  // A default (copy + in-place forget) is provided so that domains
+  // maintained OUTSIDE this repository (e.g. the PPLite native wrapper)
+  // keep compiling; in-tree domains opt in explicitly via the
+  // DEFAULT_MAKE_FORGET macro or provide a native implementation.
+  virtual Dom make_forget(const variable_vector_t &variables) const {
```
caballa, line 404: Don't do this. Intentionally, `abstract_domain.hpp` does not provide default implementations. 
Instead, crab forces to each abstract domain to explicitly define all operations. To alleviate that pain, crab provides the `DEFAULT_*` macros. 

For, the `native PPLite` domain, you need to open another PR here https://github.com/seahorn/crab-pplite. I know it's a bit painful but for licenses reasons, the `native PPLite` domain lives in a different repository.  Let me know if it's not clear and I can help

include/crab/domains/abstract_domain_macros.def
```diff
@@ -297,6 +297,25 @@ virtual void weak_assign_bool_var(const variable_t &lhs, const variable_t &rhs,
    }                                                                         \
 }
 
+/* Default implementations of make_projection/make_forget: copy the
+   whole state, then project/forget in place. Domains whose
+   representation allows building the restricted state directly (e.g.
+   environments, DBMs, apron/elina) should implement these natively. */
+#define DEFAULT_MAKE_PROJECTION(DOM)                                          \
```
caballa, line 304: Update the comments at the beginning of this file to include these two new macros

caballa (changes requested):
it looks good to me. We just need to resolve the problem with `native PPLite`.

include/crab/domains/abstract_domain.hpp
caballa: Actually after a second thought. This is okay.

caballa:
Superseded by #79.

PR #75, opened 2026-08-22
Tag analysis: source-location tags, untainted-by-default reads, and sanitizer intrinsics

LinerSu:
This PR extends the region domain's tag (taint) analysis in three
independent, stacked steps.

### 1. Source locations on variables (`3b09ee88`)

Adds `crab::source_location` (`support/debug.hpp` + `lib/debug.cpp`) and an
optional debug-info slot on `variable` (`set_debug_info`/`get_debug_info`).
A frontend can stamp a variable with the program point it came from; unset
locations cost one empty `boost::optional`.

### 2. Tags that carry their source, and untainted-by-default (`c79be3d3`)

- `region_domain_impl::tag_with_info<Number, Location>`: a tag that
  remembers the location that created it, printed as
  `TAG_<id> [src: file:line:col]`. Ordering is a strict weak order over
  `(id, location-validity, location id)`; membership stays keyed by id, so
  frontends that build query tags without debug info keep working.
- The region domain's tag type becomes `tag_with_info`; `add_tag` stamps
  the new tag with the region variable's debug info, and a failing
  `does_not_have_tag` reports the offending sources on the log channel
  `region-tag-report` (`[Sink] <loc>` + the tag set).
- **Semantic change — absent means untainted**: tag reads go through a new
  `find_tag_or_not` (absent entry = bottom = no tags) instead of
  `m_tag_env.at` (absent = top = any tag). 27 read sites converted. This
  matches the analysis' initialization ("everything starts clean") and
  makes warnings meaningful for never-tainted regions. Related rule
  adjustments: `region_copy`/`region_cast`/`ref_gep` now UNION tags into
  the destination instead of overwriting, and a strong `ref_store` of a
  non-variable (constant) clears the region's tags.
- `separate_discrete_domain::find` (pointer-or-null lookup) supports the
  above without materializing top for absent keys.

### 3. New intrinsics: `move_tag`, `print_tags`, `remove_tag` (`3fb863e2`)

- `move_tag(rgn1, ref1, rgn2, ref2)`: `tags(rgn2) := tags(rgn1)` (used by
  clam to model memcpy-style region copies).
- `print_tags(rgn, ref)`: debug dump.
- `remove_tag(rgn, ref, TAG)`: sanitizer support. Removing a tag is a
  strong update on a summarized region, so it only happens when the
  region's reference count is at most one — the same condition `ref_store`
  uses for strong updates — and is a sound no-op otherwise. Top/bottom tag
  sets are left untouched (setting a top value would drop the map entry,
  which now reads back as untainted).

### Tests (`8739141e`)

`tests/domains/region/region-8.cc`: with untainted-by-default semantics the
original ordering (tags added before the initializing stores) let a strong
constant store wipe every tag, making all checks pass vacuously; the tags
are now added after initialization and the test yields its annotated
4 safe + 1 warning. A second CFG covers `remove_tag`: strong-update removal
proves clean, an unremoved tag still warns, and a region with refcount > 1
keeps its tag (1 safe + 2 warnings). Verified by building and running the
test with clang-14; `[Sink]` reporting checked via `--log=region-tag-report`.

Deliberately NOT included: any implicit-flow machinery, and no change to
`discrete_domain::operator<=` (the existing definition already handles
bottom/top correctly).

### DIFF
--- a/include/crab/domains/region/tags.hpp
+++ b/include/crab/domains/region/tags.hpp
@@ -31,6 +31,76 @@ public:
     return o;
   }
 }; /* end class tag */
+
+// A class with more information about the source of a tag.
+template <typename Number, typename Location>
+class tag_with_info : public indexable {
+  ikos::index_t m_id; // Tag id
+  Location m_src_loc; // location that generated the tag
+  using this_class_t = tag_with_info<Number, Location>;
+  void check_id(Number n) {
+    if (n < 0) {
+      CRAB_ERROR("Cannot use negative numbers for tags");
+    }
+    if (!n.fits_int64()) {
+      CRAB_ERROR("Too large value for a tag");
+    }
+  }
+
+public:
+  tag_with_info(Number n) : m_id(0), m_src_loc(Location()) {
+    check_id(n);
+    m_id = (int64_t)n;
+  }
+  tag_with_info(Number n, Location loc) : m_id(0), m_src_loc(loc) {
+    check_id(n);
+    m_id = (int64_t)n;
+  }
+  // Lexicographic comparison over (m_id, is_valid, location id) so
+  // that operator< is a strict weak order. An invalid location
+  // carries no meaningful id, so it is ordered before any valid one
+  // and compared as id-only otherwise.
+  bool operator<(const this_class_t &as) const {
+    if (m_id != as.m_id) {
+      return m_id < as.m_id;
+    }
+    const bool this_valid = m_src_loc.is_valid();
+    const bool other_valid = as.m_src_loc.is_valid();
+    if (this_valid != other_valid) {
+      return !this_valid;
+    }
+    if (!this_valid) {
+      // both locations are invalid: compare by tag id only
+      return false;
+    }
+    return m_src_loc.get_id() < as.m_src_loc.get_id();
+  }
+  bool operator==(const this_class_t &as) const {
+    if (m_id != as.m_id) {
+      return false;
+    }
+    const bool this_valid = m_src_loc.is_valid();
+    const bool other_valid = as.m_src_loc.is_valid();
+    if (this_valid != other_valid) {
+      return false;
+    }
+    // both invalid locations compare equal (id-only comparison)
+    return !this_valid || (m_src_loc.get_id() == as.m_src_loc.get_id());
+  }
+  virtual ikos::index_t index() const override { return m_id; }
+  void write(crab_os &o) const override {
+    o << "TAG_" << m_id;
+    if (m_src_loc.is_valid()) {
+      o << " [src: " << m_src_loc << "]";
+    } else {
+      o << " [src: <unknown>]";
+    }
+  }
+  friend crab_os &operator<<(crab_os &o, const this_class_t &as) {
+    as.write(o);
+    return o;
+  }
+}; /* end class tag_with_info */
 } // end namespace region_domain_impl
 } // end namespace domains
 } // end namespace crab
--- a/include/crab/domains/region_domain.hpp
+++ b/include/crab/domains/region_domain.hpp
@@ -156,7 +156,7 @@ private:
   using alloc_site_env_t = separate_discrete_domain<variable_t, allocation_site>;
   using allocation_sites = typename alloc_site_env_t::mapped_type;
   // Map variables to sets of tags
-  using tag_t = region_domain_impl::tag<number_t>;
+  using tag_t = region_domain_impl::tag_with_info<number_t, source_location>;
   using tag_env_t = separate_discrete_domain<variable_t, tag_t>;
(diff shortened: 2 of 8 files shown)


priyasiddharth:
For my understanding, which branch is this? If I understand correctly, crab is moving to llvm18 so new functionality should ideally target that.

caballa:
Well, crab is LLVM-independent so `dev` is the right branch. However, clam has been already ported to LLVM 18 so https://github.com/seahorn/clam/pull/109 should be against `dev18`.

include/crab/domains/region/tags.hpp
```diff
@@ -31,6 +31,76 @@ template <typename Number> class tag : public indexable {
     return o;
   }
 }; /* end class tag */
+
+// A class with more information about the source of a tag.
+template <typename Number, typename Location>
```
caballa, line 36: I don't think it's worth it to have a template parameter for Location.

include/crab/domains/region_domain.hpp
```diff
@@ -520,11 +520,19 @@ class region_domain final : public abstract_domain_api<region_domain<Params>> {
...
+    } else {
+      return tag_set::bottom();
+    }
+  }
+
   template <class RangeVars>
   void merge_tags(const variable_t &x, RangeVars vars) {
     tag_set tags = tag_set::bottom();
     for (auto const &v : vars) {
-      tags = tags | m_tag_env.at(v);
```
caballa, line 527: This is a very important change everywhere. The method `at` returns `top` if the key is found. So it was very conservative: for almost any operation if we didn't know the provenance of a variable (i.e., `at` returns top) the whole set becomes top. With your change, if the provenance of a variable is unknown then the tag set is unmodified. I guess the answer boils down whether the tag analysis is a may or must analysis. @LinerSu: any thoughts?

include/crab/domains/region_domain.hpp
```diff
@@ -1019,7 +1027,8 @@ class region_domain final : public abstract_domain_api<region_domain<Params>> {
       m_rgn_equiv_classes.add(rhs_rgn, lhs_rgn);
     }
     if (crab_domain_params_man::get().region_tag_analysis()) {
-      m_tag_env.set(lhs_rgn, m_tag_env.at(rhs_rgn));
+      m_tag_env.set(lhs_rgn,
```
caballa, line 1030: Another important change. You might seem to model the operation as a weak assign. Please add a comment about the reason for the change.

include/crab/domains/region_domain.hpp
```diff
@@ -1094,7 +1103,8 @@ class region_domain final : public abstract_domain_api<region_domain<Params>> {
       m_rgn_equiv_classes.add(src_rgn, dst_rgn);
     }
     if (crab_domain_params_man::get().region_tag_analysis()) {
-      m_tag_env.set(dst_rgn, m_tag_env.at(src_rgn));
+      m_tag_env.set(dst_rgn,
```
caballa, line 1106: same here

include/crab/domains/region_domain.hpp
```diff
@@ -1593,7 +1606,8 @@ class region_domain final : public abstract_domain_api<region_domain<Params>> {
       }
 
       if (crab_domain_params_man::get().region_tag_analysis()) {
-        m_tag_env.set(ref2, m_tag_env.at(ref1));
+        // merge_tags(ref2, offset.variables());
+        m_tag_env.set(ref2, find_tag_or_not(ref2) | find_tag_or_not(ref1));
```
caballa, line 1610: weak update here again. Comment

include/crab/support/debug.hpp
```diff
@@ -76,4 +76,31 @@ void CrabEnableWarningMsg(bool b);
 extern bool CrabSanityCheckFlag;
 void CrabEnableSanityChecks(bool b);
 
+class source_location {
```
caballa, line 79: There is already a class `debug_info` in `cfg.hpp`. `source_location` is exactly a copy of that except for `m_valid` which I think it's unused.

include/crab/types/variable.hpp
```diff
@@ -321,6 +322,25 @@ class variable : public indexable {
 
   void dump(crab::crab_os &o) const { o << _n << ":" << get_type(); }
 
+  void set_debug_info(const source_location &loc) {
```
caballa, line 325: what is the meaning of having location info associate to a variable? Please add a comment, otherwise I don't see a good reason for having this

caballa (changes requested):

include/crab/domains/region_domain.hpp
caballa: I see that there is some comment in the PR description. Some of the comments should go to the code. But still is it sound the semantic change?

